Issue #5, opened 2016-08-02
Separate project structure into business and UI parts


bonkey:

bonkey:
Too early to close :)

PR #15, opened 2016-08-03
#5: Rename AtlasCheckoutDemo to AtlasDemo

bonkey:


ahmed-shehata:
There is an issue with paths in AtlasDemo project (they are not relative)

Issue #40, opened 2016-08-03
Ask user for login only AFTER they press Buy Button


bonkey:
Flow should be:

* Pick a size
* Buy button tapped
* Login user (if not logged in)

ahmed-shehata:
Related to #45 , so will close

PR #45, opened 2016-08-03
Issue 221 - Connect to Zalando

ahmed-shehata:


bonkey:
So, can we close this one? #40

PR #70, opened 2016-08-05
Testing github status API

bonkey:
It's dummy commit to test GitHub API status update for buddybuild (in correct PR instead of #69)


bonkey:
Unfortunately notifications through personal access token are not possible for zalando-incubator by default if the user is admin, so we need to skip it for now.
Fortunately we have travis for tests now too, so buddybuild could keep building demo app, but without testing and notifying the repo.

Issue #14, opened 2016-08-03
Point all unit tests to Mock Config Service


bonkey:
All unit tests should work on AtlasMockAPI server

PR #71, opened 2016-08-05
#14 - Fixed tests to point to mock server

ahmed-shehata:
Issue #14


bonkey:
1. Is it only this one place?
2. Wouldn't be possible to share some code. Maybe in base class?
3. Don't merge so quick!

ahmed-shehata:
1. Seems like so, all tests passing (without any request outside)
2. Good idea, for the mock server setup (beforeEach)

Issue #22, opened 2016-08-03
Add AtlasCheckout(options: Options) to setup AtlasSDK within AtlasCheckout


bonkey:
AtlasCheckout should have it's own AtlasSDK instance configured same way like in #21

PR #92, opened 2016-08-08
#22: Introduced basic AtlasCheckout.configure

bonkey:
(merge to issue-21, because it's heavily based on it)
- Refactored AtlasCheckout controllers to explicitly pass instance of AtlasCheckout between them
- Dropped calls to AtlasSDK.sharedInstance in AtlasUI
- Dropped CheckoutService
- Introduced (maybe temporary, maybe not) AtlasCheckoutInstance in AtlasDemo
- Made APIClient funcs public
- Made few ad-hoc code cleanups


bonkey:
👍

Issue #96, opened 2016-08-09
Fix tests based on URLSessionMock


bonkey:
Something probably broken here:

https://github.com/zalando-incubator/atlas-ios/commit/0d5ef362593c34fa15bc907684a1e233eba2165e#diff-d215e25a8a392e3f1894bdc12da2aad7

PR #100, opened 2016-08-10
#96: Added handling case when there’s no view controller to present login

bonkey:
- Added extra completion block in askUserToLogin
- Added extra completion block in execute when there’s no topViewController


bonkey:
+1

bonkey:
👍

ahmed-shehata:
👍

PR #103, opened 2016-08-10
Removing internal link

matosf:
Using a generic URL instead


matosf:
:+1:

artdaw:
👍

bonkey:
@artdaw I don't think it's that needed, but maybe we should kill it with fire this way?
https://help.github.com/articles/remove-sensitive-data/

Issue #73, opened 2016-08-05
Refactor CheckoutSummaryViewController


ahmed-shehata:
- Clean up and modulate
- Use UICollectionView instead of UIStackVIew

PR #104, opened 2016-08-10
Issue 73 - Basic refactor of CheckoutSummary + Added styler

ahmed-shehata:
Issue #73 - Basic refactor of CheckoutSummary + Added styler


AtlasUI/AtlasUI/Utils/UIView.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+extension UIView {
+    func removeAllSubviews() {
+        for view in self.subviews {
```
bonkey: Wouldn't be enough to just call `self.subviews.removeAll()` without any extension?

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -46,19 +47,21 @@ final class CheckoutSummaryViewController: UIViewController, CheckoutProviderTyp
             self.setupBlurView()
             self.setupProductImageView()
             self.setupViewLabels()
-            self.setupStackView()
+            self.view.addSubview(self.stackView)
```
bonkey: You have separated method for `setupProductImageView` with just addSubview, so make the same here

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -258,49 +173,7 @@ final class CheckoutSummaryViewController: UIViewController, CheckoutProviderTyp
...
-            stackView.addSubview(topSeparatorView)
-            topSeparatorView.heightAnchor.constraintEqualToConstant(1).active = true
-            topSeparatorView.widthAnchor.constraintEqualToAnchor(stackView.widthAnchor, multiplier: 0.95).active = true
-            topSeparatorView.bottomAnchor.constraintEqualToAnchor(stackView.topAnchor).active = true
-            topSeparatorView.centerXAnchor.constraintEqualToAnchor(stackView.centerXAnchor).active = true
-        }
-    }
-
-    private func shippingView(text: String) -> UIView? {
+    internal func shippingView(text: String) -> UIView? {
```
bonkey: It's used these functions only in Styler, so you should move it there, as there's no connection with controller.

But actually there should be connection, as you shouldn't put view hierarchy logic into styler, while you do have addSubview there.

So keep them here, but move addSubviews here as well.

bonkey:
Few minor comments, otherwise good stuff.

bonkey:
👍

AtlasUI/AtlasUI/CheckoutSummaryStyler.swift
```diff
@@ -0,0 +1,154 @@
...
+
+    private func stylizeBuyButton() {
+        viewController.buyButton.translatesAutoresizingMaskIntoConstraints = false
+
+        viewController.buyButton.heightAnchor.constraintEqualToConstant(50).active = true
+        viewController.buyButton.topAnchor.constraintEqualToAnchor(viewController.termsAndConditionsButton.bottomAnchor,
+            constant: -80).active = true
+        viewController.buyButton.leadingAnchor.constraintEqualToAnchor(viewController.view.leadingAnchor, constant: 10).active = true
+        viewController.buyButton.trailingAnchor.constraintEqualToAnchor(viewController.view.trailingAnchor, constant: -10).active = true
+        let title = viewController.customer != nil ? "Buy Now" : "Connect To Zalando"
```
bonkey: They should be two explicit and different buttons, not some extra conditionals here and there for actions / styling.

AtlasUI/AtlasUI/Utils/UIView.swift
ahmed-shehata: .subviews is read-only

bonkey:
👍

ahmed-shehata:
👍

PR #111, opened 2016-08-12
#110: Added conditional /customer in AtlasMockAPI

bonkey:
- Added conditional response from /customer based on “Authorized:” header: accepts any token
- Added AtlasMockAPIRunner – dumb-simple macOS command-line app to test AtlasMockAPI manually
  - Added swifter macOS library compilation in deps.rb
  - Added AtlasMockAPI_macOS target


bonkey:
👍

AtlasMockAPI/AtlasMockAPI/HttpServer.swift
```diff
@@ -12,6 +12,8 @@ enum HttpServerError: ErrorType {
 
 extension HttpServer {
 
+    var serverBundle: NSBundle { return NSBundle(forClass: AtlasMockAPI.self) }
```
hani-ibrahim: Why returning a new instance each time it is used?
`let serverBundle = NSBundle(forClass: AtlasMockAPI.self)`?

bonkey: that's extension... can't have properties.

hani-ibrahim: Ah yeah true :+1: 
Didn't notice that 😃

PR #127, opened 2016-08-16
CheckoutSummary code refactor

ahmed-shehata:
Refactored and modulated some of the CheckoutSummaryViewController and Styler, with the help of Mr @bonkey


ahmed-shehata:
👍

AtlasUI/AtlasUI/CheckoutSummaryStyler.swift
```diff
@@ -24,8 +24,11 @@ class CheckoutSummaryStyler: LocalizerProviderType {
         stylizeViewLabels()
         stylizeTermsButton()
 
-        stylizeBuyButton()
-        stylizeConnectToZalandoButton()
+        if Atlas.isUserLoggedIn() {
```
bonkey: Is it needed? Why not stylize always?

ahmed-shehata: Stylize "shows" the hidden button, so it would overlap with the other button, etc..

AtlasUI/AtlasUI/SizeListSelectionViewController.swift
```diff
@@ -79,13 +78,15 @@ final class SizeListSelectionViewController: UITableViewController, CheckoutProv
                 self.checkout.createCheckout(withArticle: self.article, articleUnitIndex: indexPath.row) { result in
                     switch result {
```
bonkey: Why not once here? `spinner.stopAnimating()`

AtlasUI/AtlasUI/CheckoutSummaryStyler.swift
bonkey: How about it won't "show" it?

Issue #80, opened 2016-08-05
Publish SDK to Cocoapods


artdaw:
It's time to publish SDK to the Cocoapods

bonkey:
Related #85

PR #128, opened 2016-08-16
#80: Updated CocoaPods badges

bonkey:
#80: Updated CocoaPods badges


hani-ibrahim:
👍

bonkey:
It's pity they still don't work... Will look over.

PR #141, opened 2016-08-17
enhance size list

hani-ibrahim:
Enhance the dismissing of size list in case it is one size and user is not logged in


hani-ibrahim:
👍

bonkey:
👍

bonkey:
👍

Issue #119, opened 2016-08-15
Handle if user does not have default shipping/billing addresses


ahmed-shehata:
Handle if the logged in user does not have any shipping/billing address registered, when creating cart/checkout, users should not see this screen:

![image](https://cloud.githubusercontent.com/assets/4982099/17662079/5b9ad6aa-62e3-11e6-9785-9479bd8ade01.png)

Instead, the user should have the option to add/modify shipping and billing addresses

PR #155, opened 2016-08-22
Issue 119 - Handle if user does not have default shipping/billing addresses

ahmed-shehata:
Issue #119 - Handle if user does not have default shipping/billing addresses

Notes:
- User will still be able to see the checkout screen (passed by an empty checkout object) instead of throwing an error (when no addresses are found)

- Address list is now checked before creating the checkout object, if empty will pass an empty checkout object (similar to the one when logged out)


bonkey:
👍

hani-ibrahim:
👍

bonkey:
@ahmed-shehata where are the tests? ;-)

bonkey:
👍

ahmed-shehata:
👍

Issue #151, opened 2016-08-19
Checkout Summary new design


hani-ibrahim:
Implementing the Checkout Summary new design
![summary connected](https://cloud.githubusercontent.com/assets/5626418/17810634/95272f52-661e-11e6-89ae-f6adba829937.png)

PR #173, opened 2016-08-30
Issue 151 - fixing bugs

hani-ibrahim:
follow up for issue #151 
- fixing tapping the payment selection after the order is placed
- trimming the address 
- Adding needed framework to run the app on the device


hani-ibrahim:
👍

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -91,13 +91,17 @@ class CheckoutSummaryStoryboardViewController: UIViewController, CheckoutProvide
 
     @IBAction private func shippingAddressTapped() {
         // TODO when addresses are implemented
+        guard viewState.showDetailArrow else { return }
     }
 
     @IBAction private func billingAddressTapped() {
         // TODO when addresses are implemented
+        guard viewState.showDetailArrow else { return }
```
bonkey: Seems like these two methods do nothing. Maybe in `UserMessage`

```
func notImplemented() {
  AtlasLogger.logError(...)
  show(...)
}
```

and call `userMessage.notImplemented()` here or something like this?

AtlasUI/AtlasUI/Utils/String.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+extension String {
+
+    var trimmedString: String {
```
bonkey: Type in name is redundant, just `trimmed`

hani-ibrahim: I guess It's a code style for swift
checking String properties, there are `lowercaseString` and `uppercaseString` 
what do you think?

bonkey: For ancient Swift, prior to magnificent version 3 :-)

https://github.com/apple/swift/blob/swift-3.0-branch/stdlib/public/core/String.swift#L1040

It doesn't conflict, reads better and it's future. Let's go modern.

hani-ibrahim: Good catch 👍

hani-ibrahim:
👍

ahmed-shehata:
👍

Issue #150, opened 2016-08-19
Describe installation of the SDK


bonkey:

bonkey:
Related #161

artdaw:
Done in #26

Issue #24, opened 2016-08-03
Describe new setup with callback in README.md


bonkey:

bonkey:
Related to #21 and #22

artdaw:
Done in #26

Issue #86, opened 2016-08-08
Describe initialization through Info.plist file


bonkey:
https://github.com/zalando-incubator/atlas-ios/blob/master/AtlasSDK/AtlasSDK/AppInfoConfig.swift

artdaw:
Done in #197

PR #186, opened 2016-08-31
Issue 26: fix README

artdaw:
#26, #24, #150: 3 possible installation methods added. New configuration process described in README. Usage examples fixed.


README.md
```diff
@@ -1,130 +1,158 @@
...
+    }
+}
+```
 
-1. Configure Atlas SDK first in the AppDelegate:
+## Usage
 
-    ```swift
-    import AtlasSDK
+Using shared instance previously created you can interact with SDK, for example:
```
bonkey: I would not write about shared instance at all. This is why actually I dropped it from SDK. It true it is in the demo, but it's kind of hack to make it work. It should be implemented as local variable and it might be mentioned that if the user wants to - it could be kept as shared, but it's discouraged.

BTW. with simplification of configuration, which is mentioned in #86 would allow to just call

```
Atlas.configure { client in
// ...
}
```

if Info.plist contains the correct information. And actually AtlasCheckout should have such feature, too: #187.

bonkey: So, I would also remove all `AtlasCheckoutInstance?` from code samples.

hani-ibrahim:
👍

README.md
artdaw: Examples fixed. I would add information about configuration using Info.plist later when it is done.

ahmed-shehata:
👍

artdaw:
👍

Issue #189, opened 2016-08-31
askUserToLogin: no Login WebView in case of navigation controller absence.


artdaw:
If customer has a ViewController not wrapped into any kind of NavigationController Login WebView will not open, cause `sourceApp.navigationController?` is nil, see https://github.com/zalando-incubator/atlas-ios/blob/8d8d0c5ad3822700845bcc160db8d1486e360bd7/AtlasSDK/AtlasSDK/APIRequestBuilder.swift#L73. 
I'd propose to change the line of code mentioned to the following:

`sourceApp.presentViewController(navigationController, animated: true, completion: nil)`

bonkey:
Shouldn't it just use `UIApplication.topViewController()`?

bonkey:
Ah, `sourceApp` **is** `UIApplication.topViewController()`. Should be renamed and then used like you suggested.

PR #191, opened 2016-08-31
Issue 189: askUserToLogin: no Login WebView in case of navigation controller absence

artdaw:
#189: unnecessary call to navigation controller removed


bonkey:
👍

hani-ibrahim:
👍

Issue #193, opened 2016-08-31
Allow user to select predefined shipping/billing address


ahmed-shehata:
- Allow user to select predefined shipping/billing address (when they already have some in their AddressBuch)
- Update checkout object accordingly (PUT)
- Refresh view

bonkey:
Related: #200

PR #199, opened 2016-09-02
Issue 193 - Allow user to select predefined Shipping/Billing Address

ahmed-shehata:
#193 - Allow user to select predefined Shipping/Billing Address


ahmed-shehata:
👍

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    override func tableView(tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return addresses.count
+    }
+
+    override func tableView(tableView: UITableView, cellForRowAtIndexPath indexPath: NSIndexPath) -> UITableViewCell {
+        guard let cell = tableView.dequeueReusableCellWithIdentifier(
+            String(AddressRowViewCell), forIndexPath: indexPath) as? AddressRowViewCell else {
+                fatalError("Failed to dequeue an AddressRowViewCell")
```
artdaw: I think it is not a good idea to throw fatal error, remember we have removed every fatalError call?

bonkey: Second on this.

AtlasSDK/AtlasSDK/Models/UpdateCheckoutRequest.swift
```diff
@@ -19,12 +19,19 @@ extension UpdateCheckoutRequest: JSONRepresentable {
     func toJSON() -> [String: AnyObject] {
         var result = [String: AnyObject]()
 
-        if let billingAddressId = billingAddressId {
+        if let billingAddressId = billingAddressId where !billingAddressId.isEmpty {
             result[Keys.billingAddressId] = billingAddressId
         }
-        if let shippingAddressId = shippingAddressId {
+        if let shippingAddressId = shippingAddressId where !shippingAddressId.isEmpty {
```
bonkey: You know I don't like conditions. How about extension to dictionary similar to `flatMap`, which filters out either nil or/and isEmpty values?

Then you could type something like:

`return [Keys.billingAddressId: billingAddressId, Keys.shippingAddressId: shippingAddressId].flatted(keepEmpty: false)`

I feel like it would be useful in few other places.

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,70 @@
+import Foundation
+import AtlasSDK
+
+final class AddressListViewController: UITableViewController {
+
```
bonkey: Who likes `UITableViewController`? Raise your hand! (cc: @hani-ibrahim)

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+    var delegate: AddressListViewControllerDelegate?
+
+    private let addresses: [Address]
+
+    private var selectedAddress: Address? {
+        didSet {
+            if let selectedAddress = selectedAddress {
+                dispatch_async(dispatch_get_main_queue()) {
```
bonkey: `Async.main {`

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+
+final class AddressListViewController: UITableViewController {
+
+    var delegate: AddressListViewControllerDelegate?
+
+    private let addresses: [Address]
+
+    private var selectedAddress: Address? {
+        didSet {
+            if let selectedAddress = selectedAddress {
```
bonkey: `guard let...`

AtlasSDK/AtlasSDK/Models/ShippingAddress.swift
```diff
@@ -54,3 +54,19 @@ extension ShippingAddress: JSONInitializable {
             pickupPoint: pickupPoint)
     }
 }
+
+extension ShippingAddress {
+
+    public init?(address: Address) {
```
bonkey: You might add

```
extension Addressable {
func fill(from address: Addressable) {
}
```

which could prevent from repeating most of this code in Shipping and Billing.

AtlasSDK/AtlasSDK/Models/UpdateCheckoutRequest.swift
```diff
@@ -19,12 +19,19 @@ extension UpdateCheckoutRequest: JSONRepresentable {
...
             result[Keys.shippingAddressId] = shippingAddressId
         }
         return result
     }
 }
+
+extension UpdateCheckoutRequest {
+    public init(billingAddressId: String, shippingAddressId: String) {
```
bonkey: This is redundant. Same `init` is available for every struct if there's no explicit inits defined.

AtlasUI/AtlasUI/AddressListViewController.swift
ahmed-shehata: @haldun does 😛

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
```diff
@@ -17,7 +17,27 @@ extension CheckoutSummaryActionsHandler {
...
 
         viewController.showLoader()
-        viewController.checkout.client.createOrder(checkout.id) { result in
+        if let billingAddress = viewController.checkoutViewModel.selectedBillingAddress,
+            shippingAddress = viewController.checkoutViewModel.selectedShippingAddress where
+        shippingAddress == checkout.shippingAddress &&
+        billingAddress == checkout.billingAddress {
+            createOrder(checkout.id)
+        } else {
+            let updateCheckoutRequest = UpdateCheckoutRequest(billingAddressId: viewController.checkoutViewModel.selectedBillingAddressId, shippingAddressId: viewController.checkoutViewModel.selectedShippingAddressId)
```
bonkey: Better to create

```
struct UpdateCheckoutRequest {
  init(checkoutViewModel: CheckoutViewModel) {}
}
```

instead.

bonkey: Or the best – both. One with both addresses, another as `convenience init`

AtlasSDK/AtlasSDK/Models/ShippingAddress.swift
ahmed-shehata: Shipping/Billing have some differences, where would they be addressed? (eg: pickup point,etc..)

AtlasSDK/AtlasSDK/Models/UpdateCheckoutRequest.swift
ahmed-shehata: Only for internal

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
ahmed-shehata: They are in different namespaces (AtlasUI, AtlasSDK), is it still a good idea?

AtlasSDK/AtlasSDK/Models/ShippingAddress.swift
bonkey: The differences would stay here, and they're rather small, right?

AtlasSDK/AtlasSDK/Models/CheckoutType.swift
```diff
@@ -4,6 +4,12 @@
 
 import Foundation
 
+enum CheckoutCompletness {
+    case Address
+    case Payment
+    case Valid
+}
```
bonkey: Hard to guess from the code, what actually is completeness, is it inclusive, dependent or what. I wonder if it wouldn't be better to implement 3 functions: `hasAddress`, `hasPayment` and `isValid` or something like this.

AtlasUI/AtlasUI/AddressPickerViewController.swift
```diff
@@ -97,144 +90,11 @@ final class AddressPickerViewController: UIViewController, CheckoutProviderType
...
 
 extension AddressPickerViewController: AddressListViewControllerDelegate {
-    private func addressListViewController(viewController: AddressListViewController, didSelectAddress address: Address) {
+    internal func addressListViewController(viewController: AddressListViewController, didSelectAddress address: Address) {
         delegate?.addressPickerViewController(self, pickedAddress: address, forAddressType: addressType)
     }
 }
-
-private protocol AddressListViewControllerDelegate: class {
+protocol AddressListViewControllerDelegate: class {
```
bonkey: Big antagonist of delegates, especially with single function. Just create a closure for it and done. No need to introduce additional type. Same for `AddressPickerViewControllerDelegate`

AtlasUI/AtlasUI/AddressRowViewCell.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+import AtlasSDK
+final class AddressRowViewCell: UITableViewCell {
```
bonkey: Empty line above, please.

AtlasUI/AtlasUI/AddressRowViewCell.swift
```diff
@@ -0,0 +1,69 @@
...
+    override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        setupViews()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private struct Metrics {
```
bonkey: How about using `UIEdgeInsets` instead new type?

AtlasUI/AtlasUI/AddressRowViewCell.swift
```diff
@@ -0,0 +1,69 @@
...
+        titleLabel.text = address.firstName
+
+        var lines = [String]()
+
+        if let street = address.street {
+            lines.append(street)
+        }
+
+        lines.append("\(address.zip) \(address.city)")
+        lines.append(address.countryCode)
```
bonkey: 1. `lines` should have more meaningful name
2. `lines` could be removed: `streetLabel.text = [street, "\(address.zip) \(address.city)", address.countryCode].flatMap { $0 }..joinWithSeparator("\n")
3. You should use `address.fullAddress`, and don't format it on your own

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
```diff
@@ -17,7 +17,27 @@ extension CheckoutSummaryActionsHandler {
         guard let checkout = viewController.checkoutViewModel.checkout else { return }
 
         viewController.showLoader()
-        viewController.checkout.client.createOrder(checkout.id) { result in
+        if let billingAddress = viewController.checkoutViewModel.selectedBillingAddress,
```
bonkey: Total parse error in my mind. Then confusion. What's going on here? Please extract it in meaningful chunks of code, do early return (either with `if` or `guard`) and create some meaningful syntactic code sugar, as `convenience init` or so.

AtlasUI/AtlasUI/AddressListViewController.swift
hani-ibrahim: I hate it :(
`UIViewController` with `UITableView` inside would be much better

in `UITableViewController` you can't add any view over the tableView for example the loader

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import AtlasSDK
+
+final class AddressListViewController: UITableViewController {
+
+    var delegate: AddressListViewControllerDelegate?
```
hani-ibrahim: weak ?

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
bonkey: This if could be for example:

```
if checkout.hasSameAddresses(as: viewController.checkoutViewModel) {
  createOrder(checkout.id)
  return
}
```

AtlasSDK/AtlasSDK/Models/CheckoutType.swift
ahmed-shehata: This file is deprecated, will be deleted

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+        return addresses.count
+    }
+
+    override func tableView(tableView: UITableView, cellForRowAtIndexPath indexPath: NSIndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCellWithIdentifier(
+            String(AddressRowViewCell), forIndexPath: indexPath) as! AddressRowViewCell
+        let address = addresses[indexPath.item]
+        cell.address = address
+
+        if let selectedAddress = selectedAddress where selectedAddress == address {
```
hani-ibrahim: only `if selectedAddress == address {` is enough 
especially `let selectedAddress` is not used

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
bonkey: It wouldn't hurt if you would add

```
struct CheckoutSummaryActionsHandler {

    internal unowned let viewController: CheckoutSummaryStoryboardViewController
    private var client: APIClient { return viewController.checkout.client }
    private var checkoutViewModel: CheckoutViewModel { return viewController.checkoutViewModel }


}
```

And use them instead, however it's still just hiding [feature envy anti-pattern](https://elearning.industriallogic.com/gh/submit?Action=PageAction&album=recognizingSmells&path=recognizingSmells/featureEnvy/featureEnvyExample&devLanguage=Java)

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
bonkey: Very good: one generic in SDK, one specific in UI.

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -166,3 +172,41 @@ extension CheckoutSummaryStoryboardViewController {
...
+                self.checkoutViewModel.selectedBillingAddressId = address.id
+            case AddressPickerViewController.AddressType.shipping:
+                self.checkoutViewModel.selectedShippingAddress = ShippingAddress(address: address)
+                self.checkoutViewModel.selectedShippingAddressId = address.id
+            }
+            refreshView()
+
+            if checkoutViewModel.checkout == nil &&
+            !checkoutViewModel.selectedBillingAddressId.isEmpty &&
+            !checkoutViewModel.selectedShippingAddressId.isEmpty {
```
bonkey: Meaningfullu named extension in CheckoutViewModel for this condition please...

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+        var indexPathsToReload = [indexPath]
+
+        if let selectedAddress = selectedAddress, index = addresses.indexOf(selectedAddress) {
+            indexPathsToReload.append(NSIndexPath(forItem: index, inSection: 0))
+        }
+
+        let address = addresses[indexPath.item]
+        selectedAddress = address
+        tableView.deselectRowAtIndexPath(indexPath, animated: true)
+        tableView.reloadRowsAtIndexPaths(indexPathsToReload, withRowAnimation: .Automatic)
```
hani-ibrahim: why not just `tableView.reloadData()`? and avoid all this logic ?

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -166,3 +172,41 @@ extension CheckoutSummaryStoryboardViewController {
     }
 
 }
+
+extension CheckoutSummaryStoryboardViewController {
+    func addressPickerViewController(viewController: AddressPickerViewController,
+        pickedAddress address: Address,
+        forAddressType addressType: AddressPickerViewController.AddressType) {
```
bonkey: Would be good to kill this `AddressPickerViewController.AddressType` and you would reach it using simple closures for both cases instead of Delegate Monster

AtlasUI/AtlasUI/AddressListViewController.swift
ahmed-shehata: Why reload the whole tableView for just one cell?

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -166,3 +172,41 @@ extension CheckoutSummaryStoryboardViewController {
...
+            !checkoutViewModel.selectedShippingAddressId.isEmpty {
+                showLoader()
+                checkout.client.createCheckout(checkoutViewModel.cartId, billingAddressId: checkoutViewModel.selectedBillingAddressId, shippingAddressId: checkoutViewModel.selectedShippingAddressId) { result in
+                    self.hideLoader()
+                    switch result {
+
+                    case .failure(let error):
+                        self.dismissViewControllerAnimated(true) {
+                            self.userMessage.show(error: error)
+                        }
```
bonkey: I think I saw this dismiss block few times. How about adding `dismiss: Bool` inside `UserMessage.show()`?

AtlasUI/AtlasUI/CheckoutViewModel.swift
```diff
@@ -10,15 +10,20 @@ struct CheckoutViewModel {
     let article: Article
     let selectedUnitIndex: Int
     let shippingPrice: Article.Price?
+    var cartId: String = ""
```
bonkey: Maybe it could be immutable `let`? It's passed in `init` anyway.

AtlasUI/AtlasUI/AddressListViewController.swift
hani-ibrahim: it's a tradeoff between simplicity and efficiency 
this table will not have more than 10 items for example, so reloading the whole table is not a problem
but the code will be much simpler  ;)
@bonkey what do you think?

AtlasUI/AtlasUI/CheckoutViewModel.swift
```diff
@@ -10,15 +10,20 @@ struct CheckoutViewModel {
...
-    let checkout: Checkout?
+    var checkout: Checkout?
     internal(set) var customer: Customer?
 
-    private let selectedBillingAddress: BillingAddress?
-    private let selectedShippingAddress: ShippingAddress?
+    var selectedBillingAddress: BillingAddress?
+    var selectedShippingAddress: ShippingAddress?
+
+    var selectedBillingAddressId: String = ""
```
bonkey: Don't lie ;-)

I would make both optional  instead of `""` – when they are optional

AtlasUI/AtlasUI/AddressListViewController.swift
ahmed-shehata: @haldun

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
ahmed-shehata: But then you would have to pass the dismissible view controller?

AtlasUI/AtlasUI/AddressRowViewCell.swift
```diff
@@ -0,0 +1,52 @@
...
+        streetLabel.topAnchor.constraintEqualToAnchor(titleLabel.bottomAnchor).active = true
+        streetLabel.leadingAnchor.constraintEqualToAnchor(titleLabel.leadingAnchor).active = true
+        streetLabel.trailingAnchor.constraintEqualToAnchor(titleLabel.trailingAnchor).active = true
+        streetLabel.bottomAnchor.constraintEqualToAnchor(contentView.bottomAnchor, constant: -10).active = true
+    }
+
+    private func updateViews() {
+        titleLabel.text = address.firstName
+        if let street = address.street {
+            streetLabel.text = [street, "\(address.zip) \(address.city)", address.countryCode].flatMap { $0 }.joinWithSeparator("\n")
```
hani-ibrahim: what if `address.zip` or `address.city` are empty?

ahmed-shehata: They can't be. 
Please check the docs https://atlas-checkout-api-staging.dc.zalan.do/webjars/swagger-ui/index.html?url=/api#!/address/getAddresses

hani-ibrahim: okay, but why no using `formattedAddress` property in `Addressable` protocol?

hani-ibrahim: or create another one with the format you want, to make all formatting in one place

bonkey: You wouldn't have such questions if `.fullAddress` be in use ;-)

AtlasUI/AtlasUI/AddressPickerViewController.swift
ahmed-shehata: Convince me 😛

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -166,3 +172,41 @@ extension CheckoutSummaryStoryboardViewController {
...
+
+extension CheckoutSummaryStoryboardViewController {
+    func addressPickerViewController(viewController: AddressPickerViewController,
+        pickedAddress address: Address,
+        forAddressType addressType: AddressPickerViewController.AddressType) {
+
+            switch addressType {
+            case AddressPickerViewController.AddressType.billing:
+                self.checkoutViewModel.selectedBillingAddress = BillingAddress(address: address)
+                self.checkoutViewModel.selectedBillingAddressId = address.id
```
hani-ibrahim: why we are saving the id in another property?
can't we use `self.checkoutViewModel.selectedBillingAddress.id` ?

-- I have checked that `BillingAddress` doesn't have an id and so do `Addressable` protocol ... why?
@bonkey

ahmed-shehata: backend conformity ....

ahmed-shehata: I would love to have it inside, but 😏

AtlasUI/AtlasUI/AddressListViewController.swift
bonkey: How about result-foo-shmoo-driven-cell-dequeueing?
https://github.com/zalando-incubator/atlas-ios/pull/202

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
bonkey: Well, we need to sit over on this... It should be there IMHO

AtlasUI/AtlasUI/AddressListViewController.swift
```diff
@@ -1,10 +1,10 @@
 import Foundation
 import AtlasSDK
 
-final class AddressListViewController: UITableViewController {
+final class AddressListViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
```
bonkey: Nein, nein, nein.

You can't, just can't make any class doing more than one thing.

Please make separated implementations for TVD and TVDS. I'm not joking.

AtlasUI/AtlasUI/CheckoutSummaryStoryboardViewController.swift
```diff
@@ -188,11 +188,10 @@ extension CheckoutSummaryStoryboardViewController {
             }
             refreshView()
 
-            if checkoutViewModel.checkout == nil &&
-            !checkoutViewModel.selectedBillingAddressId.isEmpty &&
-            !checkoutViewModel.selectedShippingAddressId.isEmpty {
+            if checkoutViewModel.isReadyToCreateCheckout() {
```
bonkey: At least `var readyToCreateCheckout: Bool`, but I would make it more generic in naming. However don't have better name.

ahmed-shehata:
👍

artdaw:
👍

artdaw:
👍

ahmed-shehata:
👍

Issue #223, opened 2016-09-07
Update checkout object should happen always


ahmed-shehata:
Update checkout object should happen always when user does the following scenario:

- Change shipping address
- Change billing address
- Change payment
- Issue: shipping address is back to old one

PR #224, opened 2016-09-08
#223: Added create checkout with previously selected addresses

ahmed-shehata:
#223: Added create checkout with previously selected addresses

### DIFF
--- a/AtlasUI/AtlasUI/AtlasCheckout.swift
+++ b/AtlasUI/AtlasUI/AtlasCheckout.swift
@@ -8,7 +8,7 @@ import AtlasSDK
 
 public typealias AtlasCheckoutConfigurationCompletion = AtlasResult<AtlasCheckout> -> Void
 
-typealias CreateCheckoutCompletion = AtlasResult<CheckoutViewModel> -> Void
+typealias CreateCheckoutViewModelCompletion = AtlasResult<CheckoutViewModel> -> Void
 
 public class AtlasCheckout: LocalizerProviderType {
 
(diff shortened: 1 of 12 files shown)


AtlasUI/AtlasUI/AtlasCheckout.swift
```diff
@@ -74,22 +74,25 @@ public class AtlasCheckout: LocalizerProviderType {
...
-                    let checkoutModel = CheckoutViewModel(article: article,
-                        selectedUnitIndex: selectedUnitIndex, cartId: cartId, checkout: nil)
-                    completion(.success(checkoutModel))
-                }
-
-            case .success(let checkout):
-                let checkoutModel = CheckoutViewModel(article: article,
-                    selectedUnitIndex: selectedUnitIndex, checkout: checkout)
-                completion(.success(checkoutModel))
+    func createCheckoutViewModel(withArticle article: Article, selectedUnitIndex: Int, checkoutViewModel: CheckoutViewModel? = nil,
```
hani-ibrahim: `createCheckoutViewModel` takes `CheckoutViewModel` ? 😟

ahmed-shehata: Optional one, to update if possible (billing/shipping/..)

bonkey: So maybe "prepare" instead "create". Don't lie in code, because it will lie to you one day.

bonkey: An actually why we can't use "updateCart" for this? 

I mean I can guess why, but you should ask yourself and maybe create another method, or create a common function, and use them in both: create, update and so on.

One method must do just one thing.

hani-ibrahim:
👍

artdaw:
👍

PR #230, opened 2016-09-08
#215: address is truncated and consists of 2 lines now.

artdaw:
#215: after the short discussion with @ahmed-shehata we agree that address should be truncated and consists of 2 lines.


ahmed-shehata:
👍

artdaw:
👍

Issue #221, opened 2016-09-07
raphael.koch@zalando.de on build #384: The place order button text will change for PayPal to…


bonkey:
Feedback from raphael.koch@zalando.de : The place order button text will change for PayPal to "Confirm and pay with PayPal" 

<img src="https://s3-us-west-2.amazonaws.com/buddybuild-screenshots/57a305cb34a9450100595b71/57d01090d0b94801001717b4/9d0ddff1-e1b0-4088-9db8-84d841e991ce.jpg" width="33%" height="33%" /><table><tr><td>Created</td><td>Wed Sep 07 2016 15:07:47 GMT+0000 (UTC)</td></tr><tr><td>CFBundleShortVersion</td><td>1.0</td></tr><tr><td>CFBundleVersion</td><td>20160907.384</td></tr><tr><td>App uptime</td><td>264.4025</td></tr><tr><td>Build</td><td>384</td></tr><tr><td>Device type</td><td>iPhone 6</td></tr><tr><td>Device name</td><td>Raphaels iPhone 7</td></tr><tr><td>Screen size</td><td>375</td></tr><tr><td>Screen size</td><td>375px by 667px</td></tr><tr><td>Battery</td><td>47% Unplugged</td></tr><tr><td>Memory free</td><td>54 MB / 2628 MB</td></tr><tr><td>Disk free</td><td>2119 MB / 56981 MB</td></tr><tr><td>Network IP</td><td>10.161.57.89</td></tr></table>

 [Link to buddybuild feedback from build 384](https://dashboard.buddybuild.com/apps/57a305cb34a9450100595b71/feedback?fid=57d02d43edf2d301005c22cd&bnum=384)

PR #234, opened 2016-09-09
#221: The place order button text will change for PayPal

artdaw:
Previous fix was wrong: I moved text label handling to the CheckoutViewState. Back payment also button fixed.


hani-ibrahim:
👍

artdaw:
👍

Issue #225, opened 2016-09-08
Fine-tune checkout summary screen


bonkey:
- Add delivery date
- Add colour
- Change product image to square one

PR #247, opened 2016-09-12
#225: Added Date Formatter

ahmed-shehata:
Issue #225 
- date formatter 
- doubledateformatter (thanks @bonkey) 
- added delivery label 
- removed circular image

Notes:
- Still preliminary Pull Request, yet to add tests


ahmed-shehata:
👍

AtlasUI/AtlasUI/CheckoutSummaryPriceStackView.swift
```diff
@@ -104,10 +104,11 @@ extension CheckoutSummaryPriceStackView: UIDataBuilder {
...
         vatTitleLabel.text = viewModel.loc("vat.included")
-        let estimatedDelivery = viewModel.loc("estimated.delivery") + ":"
+        let estimatedDelivery = viewModel.loc("estimated.delivery") + ": "
         var datesString = ""
-        if let earliest = viewModel.checkoutViewModel.checkout?.delivery.earliest, earliestString = viewModel.localizer.fmtDate(earliest) {
-            datesString += earliestString + "-"
+        if let earliest = viewModel.checkoutViewModel.checkout?.delivery.earliest, earliestString = viewModel.localizer.fmtDate(earliest)
+        where viewModel.checkoutViewModel.checkout?.delivery.earliest != viewModel.checkoutViewModel.checkout?.delivery.latest {
+            datesString += earliestString + " - "
         }
```
bonkey: OMG 😱

I have ¡ni puta idea! what it does. This code is super ugly.

ahmed-shehata: I thought you would comment on that as well, this is to not show both dates if they are the same, will clean 👍

AtlasUI/AtlasUI/CheckoutSummaryPriceStackView.swift
```diff
@@ -104,10 +104,11 @@ extension CheckoutSummaryPriceStackView: UIDataBuilder {
         totalTitleLabel.text = viewModel.loc("Total")
         totalValueLabel.text = viewModel.localizer.fmtPrice(viewModel.checkoutViewModel.totalPriceValue)
         vatTitleLabel.text = viewModel.loc("vat.included")
-        let estimatedDelivery = viewModel.loc("estimated.delivery") + ":"
+        let estimatedDelivery = viewModel.loc("estimated.delivery") + ": "
```
bonkey: Can't we have ": " in localized string?

AtlasUI/AtlasUI/CheckoutSummaryPriceStackView.swift
bonkey: I can guess it, but I can't understand.

AtlasUI/AtlasUI/Utils/Localization.swift
```diff
@@ -25,7 +25,7 @@ struct Localizer {
 
     var dateFormatter: NSDateFormatter {
         let dateFormatter = NSDateFormatter()
-        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS"
+        dateFormatter.dateFormat = "dd/MM/yyyy"
```
bonkey: It's not localized format at all... Use `.dateStype = .ShortStyle` or `.dateStype = .MediumStyle` instead.

AtlasUI/AtlasUI/CheckoutSummaryPriceStackView.swift
ahmed-shehata: The phrase might be needed somewhere else where ":" is not needed

AtlasUI/AtlasUI/Utils/Localization.swift
bonkey: And let's make it `let` (commit comes)

AtlasUI/AtlasUI/CheckoutSummaryPriceStackView.swift
bonkey: You shouldn't share phrases. It creates coupling.

You don't need it. You don't want it.

One place, one key.

Maybe in some super common places, like "OK", "Close" and so on we _might_ consider sharing. Not anywhere else.

bonkey:
:+1 :

ahmed-shehata:
👍

ahmed-shehata:
👍

artdaw:
👍

AtlasSDK/AtlasSDK/Models/Delivery.swift
```diff
@@ -17,7 +17,10 @@ extension Delivery: JSONInitializable {
     }
 
     init?(json: JSON) {
-        guard let latest = json[Keys.latest].string else { return nil }
-        self.init(earliest: json[Keys.earliest].string, latest: latest)
+        guard let latest = RFC3339DateFormatter().dateFromString(json[Keys.latest].string)
```
hani-ibrahim: As far as I know it's not good to allocate a new instance of `NSDateFormatter` whenever you use it.
I would prefer singleton for such things. but I know we don't use them :(
so you can at least use only one per method of class

hani-ibrahim: this apply to other files as well, like `DatesSpec.swift`

ahmed-shehata: NO singleton! We have discussed this and it's not used enough to have an effect (cc @bonkey)

hani-ibrahim: I know, that's why I said the alternative :D

bonkey: Yeah, I mentioned this to @ahmed-shehata, but it's not as bad here. We're not creating the hundreds thousands of objects anyway and this Delivery.init is used once when we get data. Optimisation here is negligible and not worth it. If you think otherwise, please provide numbers. I can bet JSON parsing here would be 100 times more costly.

Regarding singleton – we had it before for ISO8601Formatter, when it was easy to build specific object with such name. In this case I'd rather create it ad hoc.

hani-ibrahim:
👍

hani-ibrahim:
buddy build failed :(

AtlasSDK/AtlasSDK/Models/Delivery.swift
bonkey: On my machine code:

```
import Foundation

func measure(title: String, block: () -> Void) {
    let startTime = CFAbsoluteTimeGetCurrent()
    block()
    let timeElapsed = CFAbsoluteTimeGetCurrent() - startTime
    print("\(title): Time: \(timeElapsed)")
}


func formatters(count: Int) {
    var df: DateFormatter? = nil
    (1...count).forEach { _ in
        df = DateFormatter()
    }
}

[10, 100, 1000, 10000, 100000].forEach { c in
    measure(title: "\(c) times") { _ in
        formatters(count: c)
    }
}
```

Outputs:

```
10 times: Time: 0.00400996208190918
100 times: Time: 0.00262802839279175
1000 times: Time: 0.0402020215988159
10000 times: Time: 0.486325025558472
100000 times: Time: 5.15185099840164
```

bonkey: And another run with different args:

```
1 times: Time: 0.00393801927566528
2 times: Time: 0.000217020511627197
1 times: Time: 0.000122010707855225
2 times: Time: 0.000131011009216309
1 times: Time: 0.00014197826385498
2 times: Time: 0.000162005424499512
```

hani-ibrahim: 👍

ahmed-shehata:
👍

artdaw:
👍

Issue #39, opened 2016-08-03
Parse array of JSON


bonkey:
Right now we have a problem that Response accepts only JSON object as a body. There are a lot of APIs that provide array of objects, so we need somehow to deal with them.
As a workaround, for example, we have Catalog srtuct with holds an array of articles.
That should be fixed.

PR #251, opened 2016-09-14
#39: Added parsing JSON array of models

bonkey:
- Addtional fetch with [Model] argument
- Removed AddressList


bonkey:
@hani-ibrahim Maybe again my mind stuck, so if you have an idea how to merge two `fetch` funcs – you're welcome

hani-ibrahim:
this is what I thought of 
```
func fetch<Model: JSONInitializable>(from endpoint: Endpoint,
           modelCompletion: (AtlasResult<Model> -> Void)? = nil,
           arrayCompletion: (AtlasResult<[Model]> -> Void)? = nil) {

    self.requestBuilders.createBuilder(forEndpoint: endpoint, urlSession: urlSession).execute { result in
        switch result {
        case .failure(let error):
            modelCompletion?(.failure(error))
            arrayCompletion?(.failure(error))
        case .success(let response):
            if let model = Model(json: response.body) {
                modelCompletion?(.success(model))
            } else if let jsons = response.body.array.flatMap({ $0 }) {
                let models = jsons.flatMap { Model(json: $0) }
                arrayCompletion?(.success(models))
            } else {
                modelCompletion?(.failure(AtlasAPIError.invalidResponseFormat))
                arrayCompletion?(.failure(AtlasAPIError.invalidResponseFormat))
            }
        }
    }
}
```
which I know is not good 😆

bonkey:
👍

hani-ibrahim:
👍

PR #254, opened 2016-09-14
#207: Handle API responses with no body (like DELETE)

bonkey:
#207: Handle API responses with no body (like DELETE)


bonkey:
👍

AtlasSDK/AtlasSDK/APIClient.swift
```diff
@@ -17,16 +17,28 @@ public struct APIClient {
...
-    func fetch<Model: JSONInitializable>(from endpoint: Endpoint, completion: AtlasResult<Model> -> Void) {
-        fetch(from: endpoint, completion: completion) { successfulResponse in
-            return Model(json: successfulResponse.body)
+    func touch(endpoint: Endpoint, successStatus: HTTPStatus = .NoContent, completion: AtlasResult<Bool> -> Void) {
+        touch(endpoint, completion: completion) { response in
+            return response.statusCode == successStatus
+        }
+    }
+
+    func touch(endpoint: Endpoint, completion: AtlasResult<Bool> -> Void, successCompletion: JSONResponse -> Bool) {
```
hani-ibrahim: why this extra method? I think we can use `func touch(endpoint: Endpoint, successStatus: HTTPStatus = .NoContent, completion: AtlasResult<Bool> -> Void)` only to the `fetch` one
.
Or if it's a must then it could be private to avoid confusion

bonkey: Well, on purpose. Maybe you want to check something else from JSONRequest than HTTPStatus.

It's not this big, so I dared to add it this way.

hani-ibrahim:
👍

Issue #205, opened 2016-09-05
Create Add/Edit Address screen


hani-ibrahim:
Create a screen that can be used to Add and Edit the Address

PR #257, opened 2016-09-15
#205: Add/Edit Address

hani-ibrahim:
finishing #205


AtlasSDK/AtlasSDK/Models/Config.swift
```diff
@@ -14,23 +14,28 @@ public struct Config {
     let loginURL: NSURL
     let salesChannel: String
     let clientId: String
+    public let countryCode: String
```
hani-ibrahim: @bonkey, I'm not very proud of this.
Any ideas?

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Models/Config.swift
bonkey: Hm, my comment somehow disappeared...

Well, I'm facing similar problem. I did everything public so far. It's actually nothing bad.

bonkey: Well... I'm thinking now about something similar, as I'd expose `loginURL`.

How about public client.config and all properties public there as well? They're not bad actually.

bonkey: Ah, now it came, because earlier I got "You can't perform that action at this time."

hani-ibrahim: I'm not sure how these new reviews of GitHub works, I think that's why we were not able to comment

bonkey: I rather expect my problems with github.com at home when I'm not on (any) VPN

AtlasDemo/AtlasDemo/AppDelegate.swift
```diff
@@ -33,7 +33,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private var alwaysUseMockAPI: Bool {
         #if DEBUG
-            return true
+            return false
```
bonkey: Don't commit it. Default should be true

AtlasSDK/AtlasSDK/Endpoints/CheckAddressEndpoint.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+struct CheckAddressEndpoint: ConfigurableEndpoint {
+
+    let serviceURL: NSURL
+    let method: HTTPMethod = .POST
+    var path: String { return "address-checks" }
```
bonkey: `let path = "address-checks"`

AtlasSDK/AtlasSDK/Models/Config.swift
```diff
@@ -50,6 +55,7 @@ extension Config {
         self.loginURL = loginURL
         self.salesChannel = options.salesChannel
         self.clientId = options.clientId
+        self.countryCode = options.interfaceLanguage.componentsSeparatedByString("_").last ?? "DE"
```
bonkey: Should be rather opposite – you choose country and interface language is detected when not provided.

bonkey: And no defaults please.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+                    .MinLength(minLength: 3),
+                    .Pattern(pattern: "[0-9]{3,}$")]
+        case .Zipcode:
+            return [.Required,
+                    .MaxLength(maxLength: 10),
+                    .Pattern(pattern: "^[0-9]{5}$")]
+        case .City:
+            return [.Required,
+                    .MaxLength(maxLength: 50),
+                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[-,;()' 0-9"+FormValidator.anyCharacterPattern+"ß]+$")]
```
bonkey: Oh, das ist so sü**ß**! (no, it's not, it's just an irony).

Can't we have it in `anyCharacterPattern`?

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+            return [.Required,
+                    .MaxLength(maxLength: 50),
+                    .Pattern(pattern: "^(?=.*[a-zA-Z])(?=.*[0-9]).*$")]
+        case .Additional:
+            return [.MaxLength(maxLength: 50)]
+        case .Packstation:
+            return [.Required,
+                    .MinLength(minLength: 3),
+                    .MaxLength(maxLength: 3),
+                    .Pattern(pattern: "[0-9]{3,3}$")]
```
bonkey: Will be it always 3? No.

There're countries with different ids.

AtlasUI/AtlasUI/Utils/FormValidator.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+enum FormValidator {
+    case Required
+    case MinLength(minLength: Int)
+    case MaxLength(maxLength: Int)
+    case Pattern(pattern: String)
```
bonkey: You **must** provide a meaningful message to the user. Invalid value is baaaad.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
             return nil
         }
     }
+
+    var formValidators: [FormValidator] {
```
bonkey: In general: the less restrictive, the better. And all comments below.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+        switch self {
+        case .Title:
+            return [.Required]
+        case .FirstName:
+            return [.Required,
+                    .MaxLength(maxLength: 50),
+                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[- "+FormValidator.anyCharacterPattern+"ß]+$")]
```
bonkey: I would drop this pattern validation. If server respond with error - show it.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+        case .LastName:
+            return [.Required,
+                    .MaxLength(maxLength: 50),
+                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[- "+FormValidator.anyCharacterPattern+"ß]+$")]
```
bonkey: Ditto.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+        case .Street:
+            return [.Required,
+                    .MaxLength(maxLength: 50),
+                    .Pattern(pattern: "^(?=.*[a-zA-Z])(?=.*[0-9]).*$")]
```
bonkey: Ditto.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+        case .Additional:
+            return [.MaxLength(maxLength: 50)]
+        case .Packstation:
+            return [.Required,
+                    .MinLength(minLength: 3),
+                    .MaxLength(maxLength: 3),
+                    .Pattern(pattern: "[0-9]{3,3}$")]
+        case .MemberID:
+            return [.Required,
+                    .MinLength(minLength: 3),
```
bonkey: No MaxLength?

AtlasUI/AtlasUI/Utils/FormValidator.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+enum FormValidator {
+    case Required
+    case MinLength(minLength: Int)
+    case MaxLength(maxLength: Int)
```
bonkey: How about Length(min: Int, max: Int)? Would be better to get one message about it.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -96,4 +96,45 @@ enum EditAddressField {
...
+                    .Pattern(pattern: "[0-9]{3,}$")]
+        case .Zipcode:
+            return [.Required,
+                    .MaxLength(maxLength: 10),
+                    .Pattern(pattern: "^[0-9]{5}$")]
```
bonkey: Don't forget about [Limeys](https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom)

AtlasUI/AtlasUI/Utils/FormValidator.swift
```diff
@@ -0,0 +1,45 @@
...
+    case Pattern(pattern: String)
+
```
bonkey: How about `case NumbersOnly` and meaningful message with this?

AtlasUI/AtlasUI/EditAddressViewController.swift
```diff
@@ -25,15 +25,30 @@ class EditAddressViewController: UIViewController, CheckoutProviderType {
...
+        let view = LoaderView()
+        view.hidden = true
+        return view
+    }()
+
     private let addressType: EditAddressType
     internal let checkout: AtlasCheckout
     private let addressViewModel: EditAddressViewModel
-    internal var completion: EditAddressCompletion?
+    internal let completion: EditAddressCompletion
```
bonkey: `var completion: EditAddressCompletion?`?

AtlasUI/AtlasUI/EditAddressViewController.swift
```diff
@@ -55,18 +70,51 @@ extension EditAddressViewController {
...
+            let cancelButton = UIBarButtonItem(title: loc("Cancel"), style: .Plain, target: self, action: #selector(cancelButtonPressed))
+            navigationItem.leftBarButtonItem = cancelButton
+        }
     }
 
     private dynamic func submitButtonPressed() {
-        completion?(UserAddress(addressViewModel: addressViewModel))
-        dismissViewControllerAnimated(true, completion: nil)
+        let isValid = addressStackView.textFields.map { $0.validateForm() }.filter { $0 == false }.isEmpty
+        if isValid {
```
bonkey:
```
guard let request = CheckAddressRequest(addressViewModel: addressViewModel)
where isValid
else { return }
```

AtlasUI/AtlasUI/EditAddressViewModel.swift
```diff
@@ -73,22 +66,57 @@ extension UserAddress {
...
+    }
+
+}
+
+extension UpdateAddressRequest {
+
+    internal init? (addressViewModel: EditAddressViewModel) {
+
+        guard let
+            gender = addressViewModel.gender,
```
bonkey: Isn't gender optional?

AtlasUI/AtlasUI/Utils/FormValidator.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+enum FormValidator {
```
bonkey: I really like its simplicity.

AtlasUI/AtlasUI/Utils/FormValidator.swift
```diff
@@ -0,0 +1,45 @@
...
+    case MinLength(minLength: Int)
+    case MaxLength(maxLength: Int)
+    case Pattern(pattern: String)
+
+    internal func errorMessage(text: String?, localizer: LocalizerProviderType) -> String? {
+        guard !isValid(text) else { return nil }
+        return errorMessage(localizer)
+    }
+
+    internal static let anyCharacterPattern = "a-zA-ZàÀâÂäÄáÁåÅéÉèÈêÊëËìÌîÎïÏòÒôÔöÖøØùÙûÛüÜçÇñœŒæÆíóúÍÓÚĄąĆćĘęŁłŃńŚśŻżŹź"
```
bonkey: How about using `\w` matcher instead it?

```
let pattern = "\\w+"
let text = "Zażółć gęślą jaźń"
let length = text.characters.count

let regex = try? NSRegularExpression(pattern: pattern, options: .CaseInsensitive)
let matches = regex?.matchesInString(text, options: [], range: NSRange(location: 0, length: length))

matches?.forEach { match in
    let word = (text as NSString).substringWithRange(match.range)
    print("word:", word)
}
```

AtlasUI/AtlasUI/Utils/TextFieldInputStackView.swift
```diff
@@ -34,11 +34,24 @@ class TextFieldInputStackView: UIStackView {
...
         return label
     }()
 
     private weak var nextTextField: TextFieldInputStackView?
     private var valueChangedHandler: TextFieldChangedHandler?
+    private var validators: [FormValidator] = []
+    private var localizer: LocalizerProviderType!
+
+    internal func validateForm() -> Bool {
+        let error = calculateFormError()
```
bonkey: Do you count something? :)

AtlasUI/AtlasUI/Utils/UserMessage.swift
```diff
@@ -42,9 +42,9 @@ struct UserMessage {
         show(title: title, message: message, actions: ButtonAction(text: "OK"))
     }
 
-    func show(title title: String, message: String, actions: ButtonAction...) {
+    func show(title title: String, message: String?, actions: ButtonAction..., preferredStyle: UIAlertControllerStyle = .Alert) {
```
bonkey: I"d rather keep variadic arguments at the end of the arguments list.

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -59,4 +59,17 @@
...
+
+"Address.prefix.packstation" = "Packstation";
+"Address.prefix.memberID" = "Member ID";
+"Address.prefix.packstation.abbreviation" = "P.S.";
+
+"Form.validation.required" = "This field is required";
+"Form.validation.minLength: %@" = "Min length is %@";
+"Form.validation.maxLength: %@" = "Max length is %@";
+"Form.validation.pattern" = "Invalid value";
+"Address.validation.notValid" = "Address not valid, please check the address";
```
bonkey: No German copy'n'paste?

AtlasSDK/AtlasSDK/Models/Config.swift
hani-ibrahim: The requirement is not to allow the user to choose a country. That's why I needed a default country, because it's not editable.
I'll remove the default country

AtlasSDK/AtlasSDKTests/APIAddressesSpec.swift
```diff
@@ -33,7 +33,6 @@ class APIAddressesSpec: APIClientBaseSpec {
                             expect(addresses.first?.countryCode).to(equal("DE"))
                             expect(addresses.first?.isDefaultBilling).to(beTrue())
                             expect(addresses.first?.isDefaultShipping).to(beTrue())
-                            expect(addresses.first?.fullContactPostalAddress).to(equal("Erika Mustermann, Mollstr. 1\n10178 Berlin\n"))
```
hani-ibrahim: The problem is that it needs localizer now :(

AtlasSDK/AtlasSDK/Models/Config.swift
bonkey: You should pass it from options and options should have sth like:

```
        if self.authorizationHandler == nil {
            throw AtlasConfigurationError.missingAuthorizationHandler
        }
```

AtlasUI/AtlasUI/EditAddressType.swift
hani-ibrahim: but the server error will not be per field. The benefit here is that I can show instantaneous errors

AtlasSDK/AtlasSDKTests/APIAddressesSpec.swift
bonkey: Maybe mocked localizer?

AtlasUI/AtlasUI/EditAddressType.swift
hani-ibrahim: there is no max length in the mail, in the end it should be number only

AtlasUI/AtlasUI/Utils/FormValidator.swift
hani-ibrahim: not all fields have a min and max length requirement. also we show one error at a time

AtlasUI/AtlasUI/EditAddressType.swift
hani-ibrahim: which part that you want to add to `anyCharacterPattern` ?
actually this pattern consists of 2 `anyCharacterPattern` ... check the validation mail and see how the patterns looks like. I really simplified them a lot :(

AtlasUI/AtlasUI/EditAddressViewModel.swift
hani-ibrahim: no it's required in `FormattableAddress`

AtlasUI/AtlasUI/AddressListTableViewDelegate.swift
```diff
@@ -10,7 +10,8 @@ class AddressListTableViewDelegate: NSObject {
     internal var checkout: AtlasCheckout
     private let addressType: AddressType
     private let selectionCompletion: AddressSelectionCompletion
-    internal var addAddressHandler: AddAddressHandler?
+    internal var createAddressHandler: CreateAddressHandler?
```
ahmed-shehata: 👍

AtlasUI/AtlasUI/EditAddressType.swift
bonkey: Well, I would say, the more complex address form is (fields, validations, etc.), the less technologically advanced the owner/creator is.

I commented it later: how about just `\w{2,}`?

AtlasUI/AtlasUI/Utils/FormValidator.swift
bonkey: I didn't tell "instead", but additionally. But actually with one – should be better with min/max separated. Does it check it instantly along typing?

AtlasUI/AtlasUI/EditAddressType.swift
hani-ibrahim: but actually your example supports my point :D 
if we catch the error quickly by inline validation will prevent any confusions with general server error

hani-ibrahim: I'm up with any solution. but we will need to discuss it with Rafael. because in the end he is the one request inline validation

bonkey: I agree with catching it quickly. But in N26 case I'm 99% sure that somebody from frontend didn't update the rules anyway. And could be the opposite: frontend having some stricter (or wrong) rule, which is loosed on the backend and didn't allow to pass when actually it should. And SDK in app is not going to be updated...

I would love to see a backend service, which returns those rules to you. You should ask backend magicians...

AtlasUI/AtlasUI/Utils/FormValidator.swift
hani-ibrahim: It validate once the textfield is resigned active

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -117,17 +117,16 @@ enum EditAddressField {
...
             return [.Required,
                     .MinLength(minLength: 3),
-                    .Pattern(pattern: "[0-9]{3,}$")]
+                    .NumbersOnly]
         case .Zipcode:
             return [.Required,
-                    .MaxLength(maxLength: 10),
-                    .Pattern(pattern: "^[0-9]{5}$")]
+                    .ExactLength(length: 5),
+                    .NumbersOnly]
```
bonkey: Just remember it's not correct validation for UK.

bonkey: Or maybe separated issue?

bonkey:
👍

bonkey:
And standard: just merge ASAP ;-)

hani-ibrahim:
Yes I was just about to say that
- For validation issues, we have a meeting with API guys to discuss errors, this could be included in that meeting. probably they can support us with API with the patterns. To support UK, pack stations, etc.
- I have tried `\w{2,}`but it doesn't accept special characters as `-=` and spaces. or you want to replace the `anyCharacterPattern` with `\w{2,}` and leave the main pattern as it is?
- currently the UITests are failing. I'm working on it now

--
We can merge this one after fixing the failing UITests and then I have some stuff to do afterwards:
- Mocked Localizer
- Mock API for address editing
- agree on address validation
- Add UITests for Address validation

bonkey:
Then`[\w=-]{2,}` should do the job.

hani-ibrahim:
👍

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -130,7 +130,7 @@ enum EditAddressField {
         case .City:
             return [.Required,
                     .MaxLength(maxLength: 50),
-                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[-,;()' 0-9"+FormValidator.anyCharacterPattern+"ß]+$")]
+                    .Pattern(pattern: "^[,;()'0-9 \\w-]+$")]
         case .Country:
```
bonkey: All much cleaner this way!

bonkey:
👍

Issue #263, opened 2016-09-19
Adding UITests for address add/edit


hani-ibrahim:
- UITest Address creation and updating 
- Add mockApi for the addresses API

Issue #264, opened 2016-09-19
Add mocked localizer


hani-ibrahim:
Test localizer by creating mocked localizer

PR #272, opened 2016-09-20
#263: Adding UITests for Address Add/Edit screen

hani-ibrahim:
working on #263


hani-ibrahim:
👍

AtlasDemo/AtlasDemoUITests/AtlasDemoUITests.swift
```diff
@@ -80,35 +72,53 @@ class AtlasDemoUITests: XCTestCase {
...
+        app.otherElements["shipping-stack-view"].tap()
+        createAddress()
+        app.buttons["navigation-bar-cancel-button"].tap()
+    }
+
+}
+
+extension AtlasDemoUITests {
+
+    private func proceedToLogin() {
```
bonkey: I feel like it's a bit too abstract step. Or no, after second sight – not too abstract, but named incorrectly. The goal here is not to Login, but proceed buy and go to Summary _after_ login.

And also we used to differentiate between this Lamica and AFAIR Michael Kors, which was an item without sizes. We need both.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -29,6 +29,21 @@ enum EditAddressField {
...
+        case .Title: return "title-textfield"
+        case .FirstName: return "firstName-textfield"
+        case .LastName: return "lastName-textfield"
+        case .Street: return "street-textfield"
+        case .Additional: return "additional-textfield"
+        case .Packstation: return "packstation-textfield"
+        case .MemberID: return "memberID-textfield"
+        case .Zipcode: return "zipcode-textfield"
+        case .City: return "city-textfield"
+        case .Country: return "country-textfield"
```
bonkey: How about have some rest from switch-cases and have `return "\(self.lowercaseString)-textfield`"?

Less code – less maintanance. And if you already write tests – they will tell you something's wrong, you don't have to explicitly handle all enum cases.

AtlasUI/AtlasUI/EditAddressType.swift
```diff
@@ -29,6 +29,21 @@ enum EditAddressField {
...
+        }
+    }
+
     func title(localizer: LocalizerProviderType) -> String {
         switch self {
         case .Title: return localizer.loc("Address.edit.title")
```
bonkey: I strongly vote for `return "Address.edit.\(self.lowercaseString)"`?

Maybe it'll make harder automatic parse of the code to check if they're translated correctly, but you always can (and should!) write tests for all these cases instead.

hani-ibrahim:
Adding more tests and worked on #264 too

hani-ibrahim:
👍

bonkey:
👍

bonkey:
Congrats!
> All checks have passed
> 5 successful checks

And for
> codecov/project — 37.70% (+3.64%)

PR #285, opened 2016-09-21
#277: Don't display "out of stock" sizes

artdaw:
#277: unavailable (out of stock) units are removed from Article


bonkey:
👍

artdaw:
👍

bonkey:
👍

hani-ibrahim:
@bonkey @artdaw 
I'm not sure if removing the out of stock sizes is the good approach
because first It's confusing for the use as that the user my think that you don't have this size at all and second this is not the behavior in the fashion store app 
img_2273

artdaw:
@hani-ibrahim I also asked about it but AFAIK this is an agreement with Raphael

ahmed-shehata:
👍

artdaw:
👍

bonkey:
@hani-ibrahim I agree with you – would prefer to have unavailable sizes as well, but
1. It was the requirement to hide them as @artdaw told
2. Main app has "notify me when available" for them where you subscribe to notification with your email

PR #286, opened 2016-09-21
Issue 282 - Enabled pop back to summary after selecting address

ahmed-shehata:
Issue #282  - Enabled pop back to summary after selecting address

### DIFF
--- a/AtlasUI/AtlasUI/AddressListTableViewDelegate.swift
+++ b/AtlasUI/AtlasUI/AddressListTableViewDelegate.swift
@@ -14,15 +14,7 @@ class AddressListTableViewDelegate: NSObject {
     internal var updateAddressHandler: UpdateAddressHandler?
 
     var addresses: [UserAddress] = []
-    var selectedAddress: EquatableAddress? {
-        didSet {
-            Async.main { [weak self] in
-                guard let strongSelf = self, selectedAddress = strongSelf.selectedAddress else { return }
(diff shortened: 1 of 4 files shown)


AtlasUI/AtlasUI/AddressListTableViewDelegate.swift
```diff
@@ -101,6 +93,7 @@ extension AddressListTableViewDelegate: UITableViewDelegate {
             updateAddressHandler?(address: addresses[indexPath.item])
         } else {
             selectedAddress = addresses[indexPath.item]
+            self.selectionCompletion(pickedAddress: addresses[indexPath.item], pickedAddressType: self.addressType)
```
bonkey: Generally: don't put such code into `init`. It should be as simple as possible and should not fire anything.

Additionally, in the case of UIKit components – it could fire some weird behaviours.

ahmed-shehata: Not sure what you mean, it is called in the didSelectRowAtIndexPath whenever a user successfully selects an address, I have removed didSet and replaced with this

bonkey: Ach, never mind. I saw `init` just above it:

screen shot 2016-09-21 at 13 17 52

bonkey:
👍

artdaw:
👍

Issue #260, opened 2016-09-16
Extend AtlasSDK configuration


bonkey:
1. ~~Config contains list of countries with sales channels – use it instead of the one from the options~~
2. Interface language could be detected from a country and could be configured separately to it. And don't need to configure from network call.
3. ~~Use Injector instead of passing options~~ (skipped)
4. ~~Allow to modify options from Info.plist (case: read clientId from plist, change sandbox option)~~

bonkey:
Related #243

Issue #18, opened 2016-08-03
Make APIRequestBuilder UI independent


bonkey:
Meaning no LoginController if possible.

PR #288, opened 2016-09-21
#18: Big internals cleaning

bonkey:
Guys, sorry for the PR size, but when I was touching one piece, another needed some love.

- Separated login controller from AtlasSDK
- Cleaned up and simplified Options and Config code (more to be done in #260)
- Sales channel in options is deprecated and should go away when live will return correct channels
- Injector is back in the flow (for authorizationHandler)
- Removed Options from AtlasCheckout (as they're used to initialized AtlasSDK actually)


artdaw:
👍

hani-ibrahim:
I wrote these comments long time ago. Don't know that I have to approve them to appear
weird

AtlasSDK/AtlasSDK/Models/Config.swift
```diff
@@ -4,71 +4,76 @@
...
-    init(catalogAPIURL: NSURL, checkoutAPIURL: NSURL, loginURL: NSURL, countryCode: String, options: Options) {
-        self.catalogAPIURL = catalogAPIURL
-        self.checkoutAPIURL = checkoutAPIURL
-        self.loginURL = loginURL
-        self.salesChannel = options.salesChannel
-        self.clientId = options.clientId
-        self.countryCode = countryCode
+    private static func firstMatchedChannel(json: JSON, containingLocaleIdentifier identifier: String) -> [String: JSON]? {
+        let channel = json["sales-channels"].arrayValue.filter { channel in
+            return channel["locale"].stringValue.lowercaseString.containsString(identifier.lowercaseString)
```
bonkey: Well, I remember this assumption, but honestly, I don't believe in this approach. And I also remember, that nobody knows how it will be implemented in the end. That's why I would force this approach – at least for now.

Do you expect to have a list of sales channels UUIDs in a drop down in settings of the app, which uses our SDK? 👹

AtlasSDK/AtlasSDK/Options.swift
```diff
@@ -8,42 +8,50 @@ public struct Options {
...
+    public let countryCode: String
     public let configurationURL: NSURL
 
-    public init(clientId: String, salesChannel: String, useSandbox: Bool = false,
-        interfaceLanguage: String = "de_DE", configurationURL: NSURL? = nil) {
+    public var localeIdentifier: String {
+        return "\(interfaceLanguage ?? "")_\(countryCode)"
+    }
+
+    @available( *, deprecated, message = "Should be taken from config service, when ready")
```
bonkey: Well, it's "detected" based on a country and optionally language.

AtlasSDK/AtlasSDK/Utils/Injector.swift
```diff
@@ -4,39 +4,65 @@
 
 import Foundation
 
-struct Injector {
+public class Injector {
```
bonkey: In general – flexibility. Singleton is tied to given `Singleton` class. With Injector you could have different classes implementing given protocol.

bonkey:
@hani-ibrahim Actually I like "delayed" comments from reviews much better than dropping each one after another every few minutes.

AtlasDemo/AtlasDemo/ProductCollectionViewCell.swift
```diff
@@ -26,8 +26,10 @@ class ProductCollectionViewCell: UICollectionViewCell {
     }
 
     @IBAction func buyNowButtonTapped(sender: AnyObject) {
-        if let article = self.article {
-            AppSetup.sharedInstance.checkout?.presentCheckoutView(sku: article.id)
+        if let rootController = UIApplication.sharedApplication().keyWindow?.rootViewController,
```
bonkey: Yeah, thought about, but it's in a single place, so didn't want to play with some funny static shortcuts without any specific logic.

AtlasSDK/AtlasSDK/Models/Config.swift
hani-ibrahim: I don't expect that for sure. But also the current mapping is just a coincidence that every sales channel has one country, I think because we support only customer which is stylist. But when we support style fruit. then we will have another sales channel for them but with the same country. 

So I would not go with this approach and ask API how this should be handled properly. 
I'm not happy with my solution or with yours :D. Both are just hacks

hani-ibrahim:
👍

hani-ibrahim:
👍

hani-ibrahim:
👍

ahmed-shehata:
👍

bonkey:
👍

bonkey:
👍

bonkey:
👍

artdaw:
👍

Issue #64, opened 2016-08-04
Standardize Asynchronous UI changes with Network Requests to run on main thread


ahmed-shehata:
In the execute function [here](https://github.com/zalando-incubator/atlas-ios/blob/5b0c473bb6358eb6f8a60101043b69b8ece7748d/AtlasSDK/AtlasSDK/APIRequestBuilder.swift#L21), not ALL network requests should run on main thread, just the ones that will affect the UI.

This would require having a different call / flag / etc. to specify if this should be run on the main thread.

bonkey:
Especially `dispatch_async` in https://github.com/zalando-incubator/atlas-ios/blob/5b0c473bb6358eb6f8a60101043b69b8ece7748d/AtlasSDK/AtlasSDK/APIRequestBuilder.swift#L21

bonkey:
We can pass specific queue to `execute()` to run on it, with main as default

bonkey:
All callbacks done on the main thread in RequestBuilder has been removed in #18 and #260, so it should be closed for now. If there're some other issues – they should get another issue.

PR #296, opened 2016-09-21
Issue #295 Fixed showing loader in main thread

ahmed-shehata:
Issue #295: 

- Fixed showing loader in main thread
- Fixed bug when deleting preselected address
- Added ability to enable/disable pop back to summary


AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
```diff
@@ -154,9 +154,11 @@ extension CheckoutSummaryActionsHandler {
 
 extension CheckoutSummaryActionsHandler {
     func pickedAddressCompletion(pickedAddress address: EquatableAddress?,
-        forAddressType addressType: AddressType) {
+        forAddressType addressType: AddressType, popBackToSummary: Bool) {
 
-            viewController.navigationController?.popViewControllerAnimated(true)
+            if popBackToSummary {
+                viewController.navigationController?.popViewControllerAnimated(true)
```
bonkey: Does this really pop back to summary? Maybe argument should be named `popControllerOnFinish`? Still, not a big fan of this extra argument and ifelsism.

bonkey: And should it be at the beginning? Maybe at least in `defer { ... }` block?

ahmed-shehata: As far as the flow goes, it must pop first back and then continue on all the other logic (eg: refreshing the view, creating a new checkout object, etc..)

bonkey: Maybe I don't want to know what kind of things from other logic it touches. Yeah, I'm pretty sure I don't want to know ;-)

bonkey:
👍

ahmed-shehata:
👍

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -57,11 +57,15 @@ class CheckoutSummaryViewController: UIViewController, CheckoutProviderType {
     }
 
     internal func showLoader() {
-        loaderView.show()
+        Async.main {
```
hani-ibrahim: why?
I'm already doing this inside the LoaderView
```
internal func show() {
        Async.main {
            self.activityIndicator.startAnimating()
            self.hidden = false
        }
    }
```

artdaw:
👍

hani-ibrahim:
please check my comment first ;)

AtlasSDK/AtlasSDK/APIClient.swift
```diff
@@ -44,12 +44,18 @@ public struct APIClient {
...
             switch result {
             case .failure(let error):
-                completion(.failure(error))
+                dispatch_async(dispatch_get_main_queue()) {
+                    completion(.failure(error))
+                }
             case .success(let response):
                 if let parsedResponse = successHandler(response) {
-                    completion(.success(parsedResponse))
+                    dispatch_async(dispatch_get_main_queue()) {
```
hani-ibrahim: @bonkey I decided with Ahmed to move the Async.main to the SDK for the following reasons
- it does not make sense to the user of the SDK to call the SDK on the main thread and then receive the callback on a background thread. The best case is to receive the callback on the same thread that the user used to call the SDK (which could be tricky) but at least using the main thread and not an unknown background thread
- We were using an Async.main whenever we show or hide a loader. But if the two calls to show and then hide the loader are really fast then an unexpected behavior could happen because each call to show and hide will be scheduled to the main thread which might not be in the same order that we called them in the first place. Ex: calling show then hide could be scheduled to the main thread as hide then show
- We had a strange crash because the UI is not updated in the main thread which we couldn't fix expect by this solution because the debugger didn't show us the line of code where the crash occurs
- Ahmed told me that you were against that in the past because parsing the data could take some time and we should not block the UI for that. But now the Async.main is called only for the completion block which is better I think

Your opinion?

bonkey: 1. `Async` is in UI, because it's hevily used only there
2. There's only a single place where async on main thread is needed in SDK – so no need to move it to SDK and make it public for this single usage
3. I tried to fix main thread completion calls Ahmed mentioned (#64) in my last #18, but in the end it stayed as it was: https://github.com/zalando-incubator/atlas-ios/blob/e3104e05cdfdf139acd354f7b35b5ca5b0b6ed90/AtlasSDK/AtlasSDK/RequestBuilder.swift#L43

So, IMO please keep Async in UI

hani-ibrahim: Sorry I described it wrongly. I didn't move `Async.main` but used the apple one instead. `dispatch_async`. I know that it should be in UI and SDK should not use the UI framework

hani-ibrahim:
👍

ahmed-shehata:
👍

hani-ibrahim:
👍

bonkey:
👍

PR #299, opened 2016-09-22
Updated deps

bonkey:
- Tied Nimble to v4 which is swift2.x compatible
- Switched to official SwiftyJSON on “swift2” branch


Cartfile.private
```diff
@@ -1,14 +1,14 @@
 # Tests
 
 github "Quick/Quick"
-github "Quick/Nimble"
+github "Quick/Nimble" "v4.1.0"
```
hani-ibrahim: why?

bonkey:
👍

Cartfile.private
bonkey: 5.0 is swift3 only

ahmed-shehata:
👍

PR #305, opened 2016-09-23
#260: Dropped country from Options, allow to choose any language

bonkey:
#260: Dropped country from Options, allow to choose any language

### DIFF
--- a/AtlasSDK/AtlasSDK/Models/Config.swift
+++ b/AtlasSDK/AtlasSDK/Models/Config.swift
@@ -15,6 +15,10 @@ public struct Config {
     public let salesChannelLocale: NSLocale
     public let interfaceLocale: NSLocale
 
+    public var countryCode: String {
+        return interfaceLocale.countryCode
+    }
+
 }
 
 private typealias LocaleSalesChannel = (salesChannel: String, locale: String)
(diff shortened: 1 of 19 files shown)


bonkey:
👍

AtlasSDK/AtlasSDK/Models/Config.swift
```diff
@@ -12,14 +12,11 @@ public struct Config {
...
+    public let salesChannelLocale: NSLocale
+    public let interfaceLocale: NSLocale
 
     public var countryCode: String {
-        return locale.countryCode
-    }
-
-    public var languageCode: String {
-        return locale.languageCode
+        return interfaceLocale.countryCode
```
hani-ibrahim: `interfaceLocale.countryCode` or `salesChannelLocale.countryCode`?
I think it read from `salesChannelLocale`. as the user can't create an address that ship to the country based on interface language but based on the salesChannel

bonkey: Interface locale has **only** language set up from options. Country is **always** from sales channel

bonkey: See https://github.com/zalando-incubator/atlas-ios/blob/3945519a91dd243653b3ccc4ef5c24a4ac5d58d7/AtlasSDK/AtlasSDKTests/ConfigSpec.swift#L57

hani-ibrahim:
👍

PR #311, opened 2016-09-23
#309: returning the forms patterns

hani-ibrahim:
working on #309

### DIFF
--- a/AtlasUI/AtlasUI/AddressFormType.swift
+++ b/AtlasUI/AtlasUI/AddressFormType.swift
@@ -106,11 +106,11 @@ enum AddressFormField: String {
         case .FirstName:
             return [.Required,
                     .MaxLength(maxLength: 50),
-                    .Pattern(pattern: "^[' \\w-]+$")]
+                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[- "+FormValidator.anyCharacterPattern+"ß]+$")]
         case .LastName:
             return [.Required,
                     .MaxLength(maxLength: 50),
-                    .Pattern(pattern: "^[' \\w-]+$")]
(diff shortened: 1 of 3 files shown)


hani-ibrahim:
👍

AtlasUI/AtlasUI/AddressFormType.swift
```diff
@@ -132,7 +132,7 @@ enum AddressFormField: String {
         case .City:
             return [.Required,
                     .MaxLength(maxLength: 50),
-                    .Pattern(pattern: "^[,;()'0-9 \\w-]+$")]
+                    .Pattern(pattern: "^["+FormValidator.anyCharacterPattern+"]'?[-,;()' 0-9"+FormValidator.anyCharacterPattern+"ß]+$")]
```
bonkey: All those three are same, right? So at least don't DRY and create

```
enum FormValidator {
  internal static let anyNamePattern = "^[\(FormValidator.anyCharacterPattern)]'?[- \(FormValidator.anyCharacterPattern)ß]+$"
}
```

bonkey:
👍

AtlasUI/AtlasUI/AddressFormType.swift
hani-ibrahim: only the first two are the same and the third is different

bonkey: I'm not trying to be pain-in-the-butt, but it just proves it should be extracted to two separated and named patterns: name, city), as it's pretty unreadable this way and I didn't spot the difference.

hani-ibrahim: completely agree

hani-ibrahim:
👍

bonkey:
👍

PR #313, opened 2016-09-24
#312: Get rid of tens of billing/shipping id calls

bonkey:
I know it might be better to use simple string ids at least in lowest call in APIClient, but I think devs would work with existing addresses anyway.

### DIFF
--- a/AtlasUI/AtlasUI/AtlasCheckout.swift
+++ b/AtlasUI/AtlasUI/AtlasCheckout.swift
@@ -56,20 +56,22 @@ public class AtlasCheckout: LocalizerProviderType {
         viewController.presentViewController(navigationController, animated: true, completion: nil)
     }
 
-    func createCheckout(fromModel checkoutViewModel: CheckoutViewModel,
+    func createCheckoutViewModel(from checkoutViewModel: CheckoutViewModel,
         completion: CreateCheckoutViewModelCompletion) {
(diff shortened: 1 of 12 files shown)


AtlasUI/AtlasUI/AtlasCheckout.swift
```diff
@@ -56,22 +56,26 @@ public class AtlasCheckout: LocalizerProviderType {
         viewController.presentViewController(navigationController, animated: true, completion: nil)
     }
 
-    func prepareCheckoutViewModel(selectedArticleUnit: SelectedArticleUnit, checkoutViewModel: CheckoutViewModel? = nil,
+    func createCheckout(fromModel checkoutViewModel: CheckoutViewModel,
```
ahmed-shehata: Why change the function name? This does not return a checkout.

bonkey:
@ahmed-shehata forgive my blindness but for my eyes it does

AtlasUI/AtlasUI/AtlasCheckout.swift
bonkey: Ahc, now I see

bonkey:
👍

artdaw:
👍

hani-ibrahim:
👍

artdaw:
👍

PR #318, opened 2016-09-26
Updated deps

bonkey:
- Froze Quick to v0.9 - swift2 compatibility
- Switeched back to bonkey/swifter/stable as httpswift/swifter/stable is swift3 only now


bonkey:
👍

artdaw:
👍

Issue #255, opened 2016-09-14
Incorrect authorization token breaks flow


bonkey:
When authorization token is invalid and backend responds with 401 Unauthorised the flow in UI breaks.

Maybe 401 needs to be handled differently or backend should introduce 403 case?

To quickly reproduce: first login with mock service (it will create token with `TOKEN` value), then switch to real backend.

artdaw:
Things to do in this issue:

- [ ] Remove authorization token in GetArticle request cause it is not needed by API
- [ ] Fix "Checkout with Zalando button"

PR #319, opened 2016-09-26
Issue 255: bad authorisation token

artdaw:
#255: authorization token removed in GetArticle and Config request cause it is not needed by API

### DIFF
--- a/AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
+++ b/AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
@@ -14,6 +14,7 @@ protocol Endpoint: CustomStringConvertible {
     var queryItems: [NSURLQueryItem]? { get }
     var parameters: [String: AnyObject]? { get }
 
+    var isOAuth: Bool { get }
 }
 
 extension Endpoint {
@@ -26,6 +27,7 @@ extension Endpoint {
     var queryItems: [NSURLQueryItem]? { return nil }
     var parameters: [String: AnyObject]? { return nil }
 
+    var isOAuth: Bool { return true }
 }
 
 extension Endpoint {
--- a/AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
+++ b/AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
@@ -14,7 +14,7 @@ protocol Endpoint: CustomStringConvertible {
     var queryItems: [NSURLQueryItem]? { get }
     var parameters: [String: AnyObject]? { get }
 
-    var isOAuth: Bool { get }
+    var requiresAuthorization: Bool { get }
 }
 
 extension Endpoint {
@@ -27,7 +27,7 @@ extension Endpoint {
     var queryItems: [NSURLQueryItem]? { return nil }
     var parameters: [String: AnyObject]? { return nil }
 
-    var isOAuth: Bool { return true }
+    var requiresAuthorization: Bool { return true }
 }
 
 extension Endpoint {
--- a/AtlasSDK/AtlasSDK/Endpoints/GetConfigEndpoint.swift
+++ b/AtlasSDK/AtlasSDK/Endpoints/GetConfigEndpoint.swift
@@ -7,7 +7,5 @@ import Foundation
 struct GetConfigEndpoint: Endpoint {
 
     let URL: NSURL
-    var isOAuth: Bool {
-        return false
-    }
+    let requiresAuthorization = false
 }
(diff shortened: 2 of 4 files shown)


AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
```diff
@@ -14,6 +14,7 @@ protocol Endpoint: CustomStringConvertible {
     var queryItems: [NSURLQueryItem]? { get }
     var parameters: [String: AnyObject]? { get }
 
+    var isOAuth: Bool { get }
```
bonkey: I would call it more generic: `requiresAuthorization`

AtlasSDK/AtlasSDK/Endpoints/GetConfigEndpoint.swift
```diff
@@ -7,5 +7,7 @@ import Foundation
 struct GetConfigEndpoint: Endpoint {
 
     let URL: NSURL
-
+    var isOAuth: Bool {
```
bonkey: Just `let isOAuth = false`

AtlasSDK/AtlasSDK/RequestBuilder.swift
```diff
@@ -65,8 +65,8 @@ struct RequestBuilder {
     }
 
     private func buildRequest() throws -> NSMutableURLRequest {
-        let request = try NSMutableURLRequest(endpoint: endpoint)
-        return request.authorize(withToken: APIAccessToken.retrieve()).debugLog()
+        let request = try NSMutableURLRequest(endpoint: endpoint).debugLog()
```
bonkey: debugLog() is not needed here

AtlasSDK/AtlasSDK/RequestBuilder.swift
```diff
@@ -65,8 +65,8 @@ struct RequestBuilder {
...
+        return endpoint.isOAuth ? request.authorize(withToken: APIAccessToken.retrieve()).debugLog() : request.debugLog()
```
bonkey: I'd rather go with guard here:

```
let request = try NSMutableURLRequest(endpoint: endpoint)
guard endpoint.requiresAuthorization else {
  return request.debugLog()
}
return request.authorize(withToken: APIAccessToken.retrieve()).debugLog()
```

artdaw: Could you, please, explain why?

AtlasSDK/AtlasSDK/RequestBuilder.swift
artdaw: Fixed in merge

AtlasSDK/AtlasSDK/Endpoints/GetConfigEndpoint.swift
artdaw: Fixed

AtlasSDK/AtlasSDK/RequestBuilder.swift
bonkey: Because I'm old and blind. When I was young, I used to use sometimes 3 ternary operators in one line, to make it more terse, but now clarity is the most important and when I have one instruction in line and guard clearly tells it's "exceptional" situation and early exit I can understand what code does without even reading it.

bonkey: And other Swift devs are blind too, which is visible in [Swift guidelines](https://swift.org/documentation/api-design-guidelines/):
> Clarity is more important than brevity. Although Swift code can be compact, it is a non-goal to enable the smallest possible code with the fewest characters. Brevity in Swift code, where it occurs, is a side-effect of the strong type system and features that naturally reduce boilerplate.

haldun: Then, maybe `if` is better than `guard` here?

artdaw: Okaaay, fair enough. Changed it.

bonkey: Is it? I prefer to have a bouncer on top of the function to get rid of all corner cases.

haldun: It is already 2 lines of code in this function :)

bonkey: @haldun Ach, good old code-style-flame-war... But not today, I haven't even started things I wanted to do in the morning.

bonkey:
👍

artdaw:
👍

PR #322, opened 2016-09-26
Issue 304 - Generalizing errors

hani-ibrahim:
Working on #304


hani-ibrahim:
👍

AtlasUI/AtlasUI/AddressFormViewController.swift
```diff
@@ -154,32 +154,24 @@ extension AddressFormViewController {
...
-            if checkAddressResponse.status == .notCorrect {
-                let title = self.loc("Address.validation.notValid")
-                self.userMessage.show(title: title, message: nil, actions: ButtonAction(text: "OK"))
-            } else {
-                switch self.addressMode {
-                case .createAddress: self.createAddressRequest()
-                case .updateAddress(let address): self.updateAddressRequest(address)
-                }
-                self.dismissView()
+        guard let checkAddressResponse = result.handleError(checkoutProviderType: self) else { return }
```
bonkey: Everything great, but we need to get better naming for sure.

When I just looked at this I was so puzzled: why the handled error is assigned to checkAddressResponse? :)

bonkey: Anything in this direction? (from the worst to the best IMHO)

* (hopefully not) `result.getSuccess(...)`
* `result.extractSuccess(...)`
* `result.mapSuccess(...)`
* `result.takeSuccess(...)`
* `result.selectSuccess(...)`

Or maybe even just `result.success(...)`?

```
guard let checkAddress = result.success(checkoutProviderType: self) else { return }
```

AtlasUI/AtlasUI/de.lproj/Localizable.strings
```diff
@@ -75,3 +75,6 @@
 "Form.validation.pattern" = "Invalid value";
 "Form.validation.numbersOnly" = "Only numbers are allowed";
 "Address.validation.notValid" = "Address not valid, please check the address";
+
+"Error.general.title" = "Oops";
+"Error.general.message" = "Sorry, Something went wrong!";
```
bonkey: Don't! Shout! At! People! Please!

AtlasUI/AtlasUI/AddressFormViewController.swift
hani-ibrahim: I think we need to indicate that the error is handled somehow
`result.successOrHandleError(...)` ?

or
I can change
```swift
internal func success(checkoutProviderType checkoutProviderType: CheckoutProviderType, type: AtlasUIError = .GeneralError) -> T? {
```
to
```swift
internal func success(errorHandlingType errorHandling: AtlasUIError) -> T? {
```
and the `checkoutProviderType` will be inside the `AtlasUIError` as it's only needed for the error only and will be removed later after the refactor of the localization

bonkey: I thought about this "error handling indicator" yesterday and in every function name I found, it was misleading because you don't want to know anything what's happening with the error, actually. If you would, you should use former `switch result {}`

So, this default errorHandlingType is pretty good compromise, if it won't complicate things.

bonkey:
👍

hani-ibrahim:
👍

AtlasUI/AtlasUI/AddressListTableViewDelegate.swift
```diff
@@ -10,17 +10,19 @@ class AddressListTableViewDelegate: NSObject {
...
+    private let userMessage: UserMessage
     internal var createAddressHandler: CreateAddressHandler?
     internal var updateAddressHandler: UpdateAddressHandler?
     internal var deleteAddressHandler: DeleteAddressHandler?
 
     var addresses: [UserAddress] = []
     var selectedAddress: EquatableAddress?
-    init(checkout: AtlasCheckout, addressType: AddressType,
+    init(checkout: AtlasCheckout, addressType: AddressType, userMessage: UserMessage,
         addressSelectionCompletion: AddressSelectionCompletion) {
```
ahmed-shehata: why pass around UserMessage?

Issue #317, opened 2016-09-26
Add ToC link (depended on sales channel)


bonkey:

bonkey:
It's sales channel related, so add to config service.

artdaw:
Fix after https://github.bus.zalan.do/Atlas/config/issues/18 is done

PR #336, opened 2016-09-27
#317: Add ToC link

artdaw:
#317: ToC lins dependant of sales channel added

### DIFF
--- a/AtlasUI/AtlasUI/CheckoutSummaryFooterStackView.swift
+++ b/AtlasUI/AtlasUI/CheckoutSummaryFooterStackView.swift
@@ -6,13 +6,16 @@ import UIKit
 
 class CheckoutSummaryFooterStackView: UIStackView {
 
-    internal let footerLabel: UILabel = {
-        let label = UILabel()
-        label.numberOfLines = 0
-        label.textAlignment = .Center
-        label.font = .systemFontOfSize(12, weight: UIFontWeightLight)
-        label.textColor = UIColor(hex: 0xB2B2B2)
-        return label
+    var tocURL: NSURL?
+
+    internal let footerButton: UIButton = {
+        let button = UIButton(type: .System)
+        button.titleLabel?.font = .systemFontOfSize(12)
+        button.titleLabel?.lineBreakMode = .ByWordWrapping
+        button.titleLabel?.numberOfLines = 2
+        button.titleLabel?.textAlignment = .Center
+        button.accessibilityIdentifier = "checkout-summary-toc-button"
+        return button
     }()
 
     internal let submitButton: RoundedButton = {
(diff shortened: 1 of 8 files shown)


AtlasUI/AtlasUI/CheckoutSummaryFooterStackView.swift
```diff
@@ -29,19 +32,29 @@ class CheckoutSummaryFooterStackView: UIStackView {
...
-        addArrangedSubview(footerLabel)
+        footerButton.addTarget(self, action: #selector(CheckoutSummaryFooterStackView.tocPressed(_:)), forControlEvents: .TouchUpInside)
+        addArrangedSubview(footerButton)
         addArrangedSubview(submitButton)
     }
 
+    @objc func tocPressed(sender: UIButton!) {
+        if let url = tocURL {
+            let controller = ToCViewController(tocURL: url)
+            UIApplication.topViewController()?.navigationController?.pushViewController(controller, animated: true)
```
bonkey: 1. I wonder, ToC might be that important from legal perspective, that it should be somehow **always** handled in case `UIApplication.topViewController()?.navigationController?` is actually `nil` – maybe open Safari?
2. I feel like you shouldn't take `navigationController?` "from somewhere" (who knows what such static method calls), as it makes indirect assumption and dependency that this will be always there

AtlasUI/AtlasUI/CheckoutSummaryFooterStackView.swift
```diff
@@ -29,19 +32,33 @@ class CheckoutSummaryFooterStackView: UIStackView {
...
 
+    @objc func tocPressed(sender: UIButton!) {
+        if let url = tocURL {
+            let controller = ToCViewController(tocURL: url)
+            guard let navController = UIApplication.topViewController()?.navigationController else {
+                UIApplication.sharedApplication().openURL(url)
+                return
+            }
+            navController.pushViewController(controller, animated: true)
+        }
```
bonkey: Yeah, _just_ code style ;-)

```
guard let url = tocURL else { return }
guard let navController = UIApplication.topViewController()?.navigationController else {
  UIApplication.sharedApplication().openURL(url)
  return
}
navController.pushViewController(ToCViewController(tocURL: url), animated: true)
```

artdaw: Ah, you are right. Fixed.

bonkey:
👍

AtlasUI/AtlasUI/CheckoutSummaryFooterStackView.swift
```diff
@@ -29,19 +32,31 @@ class CheckoutSummaryFooterStackView: UIStackView {
...
         addArrangedSubview(submitButton)
     }
 
+    @objc func tocPressed(sender: UIButton!) {
+        guard let url = tocURL else { return }
+        guard let navController = UIApplication.topViewController()?.navigationController else {
+            UIApplication.sharedApplication().openURL(url)
+            return
+        }
+        navController.pushViewController(ToCViewController(tocURL: url), animated: true)
```
hani-ibrahim: why not displaying it inside our UI and remove this dependency at all?

artdaw: It's the same web view as for login or payment. Let's keep it like this now, maybe we migrate to SFViewController soon.

ahmed-shehata:
👍

artdaw:
👍

PR #354, opened 2016-09-28
#347: device info added to Bug Screenshot

artdaw:
#347: device info added to Bug Screenshot


bonkey:
This merge should go to `bug-bashing` branch as I wrote on hipchat.

PR #361, opened 2016-09-28
Update localizable.string with german

Gregsen:
If updated most fields. I am not sure about line 16 (CheckoutSummaryViewController.terms), because in the shop, it also mentions data protection regulations (Datenschutzbestimmung) and cancellation instruction (Wiederrufsbelehrung).
I also left some word in english (response, checkput), since they are for devs and they should know these words.

Finally, i left out member id (Address.form.memberid, Address.prefix.memberID) and pick up point , because i don't know what that would be in german


ahmed-shehata:
👍

bonkey:
I'm sorry @Gregsen, but archangel Raphael enlightened me yesterday and I translated them instantly in #364, so I'm closing this one.

Issue #338, opened 2016-09-27
Add lookback.io into Demo app


bonkey:
It's pretty easy to [manually add](https://lookback.io/learn/ios-sdk/get-started) it and required frameworks, and init with token: `yxHNcPapi4XecJDit`

It should help us to gather feedback during bug bashing sessions.

bonkey:
Questions to answer:
1. How to identify user?
2. How to highlight important part of whole recording

bonkey:
Record yourself to create simple introduction how does it work for testers

ahmed-shehata:
It's. Just. Amazing.

PR #366, opened 2016-09-29
#338: Added lookback SDK

ahmed-shehata:
#338: Added lookback SDK


bonkey:
Just for future: you should add only those files which needed. No docs, no two versions of the framework.

Issue #324, opened 2016-09-27
Checkout shouldn't reset if billing address (only) is changed


ahmed-shehata:
Scenario:
- User has a full checkout object
- User goes to delete billing address 
- Shipping address is deleted (this should not happen)

PR #370, opened 2016-09-29
#324: Fixed bug where addresses would be both resetted

ahmed-shehata:
#324: Fixed bug where addresses would be both resetted


artdaw:
👍

ahmed-shehata:
👍

Issue #327, opened 2016-09-27
Error handling: Banner inside presented controller


bonkey:
This banner should be used to show errors, like `UserMessage` does at this moment. It could just replace UIAlert implementation.

Instead of small "X" it should maybe automatically dismiss or contain "OK" button

**Note:** It could contain "Try again" button on side, but first we need to clarify how to simply retry whole operation (maybe using something like in [NSInvocation](https://developer.apple.com/reference/foundation/nsinvocation)). If only updating operation doesn't alternate screen state it's not necessary as user can interact with the same action caller (button) again

![error](https://cloud.githubusercontent.com/assets/39100/18867175/78f89b10-84a3-11e6-9c58-8744beb10fd3.jpg)

PR #385, opened 2016-09-30
#327: banner error

hani-ibrahim:
working on #327 

simulator screen shot 30 sep 2016 6 38 10 pm

### DIFF
--- a/AtlasUI/AtlasUI/Utils/UIView.swift
+++ b/AtlasUI/AtlasUI/Utils/UIView.swift
@@ -4,6 +4,23 @@
 
 import Foundation
 
+enum ViewAnchor {
+    case top
+    case right
+    case bottom
+    case left
+
+    func anchorForView(view: UIView) -> NSLayoutAnchor {
+        switch self {
+        case .top: return view.topAnchor
+        case .right: return view.rightAnchor
+        case .bottom: return view.bottomAnchor
+        case .left: return view.leftAnchor
+        }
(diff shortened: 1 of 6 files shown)


bonkey:
👍

AtlasUI/AtlasUI/Utils/UIView.swift
```diff
@@ -4,6 +4,23 @@
...
+    case right
+    case bottom
+    case left
+
+    func anchorForView(view: UIView) -> NSLayoutAnchor {
+        switch self {
+        case .top: return view.topAnchor
+        case .right: return view.rightAnchor
+        case .bottom: return view.bottomAnchor
+        case .left: return view.leftAnchor
```
bonkey: Isn't it "it looks like a nail" approach? New type, just to translate property to type and play with enum :)
Smells like unneeded piece of code.

hani-ibrahim: I was thinking about readability.
otherwise I'll need to pass both the view anchor and the superview anchor in the `snapAnchorToSuperView` method

bonkey: Well... Maybe. It also introduces new code to maintain and another code to _read_.
Not sure if better than two explicit anchors – the best code is always no code. But let's keep it.

hani-ibrahim:
👍

PR #388, opened 2016-10-04
#387: Handle network error

hani-ibrahim:
Working on #387 
img_2549

### DIFF
--- a/AtlasUI/AtlasUI/AtlasUIErrors.swift
+++ b/AtlasUI/AtlasUI/AtlasUIErrors.swift
@@ -52,12 +52,19 @@ extension UserPresentable {
 
 extension AtlasAPIError: UserPresentable {
 
+    func shouldDisplayGeneralMessage() -> Bool {
+        switch self {
+        case let .nsURLError(_, details): return details == nil
+        default: return true
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


hani-ibrahim:
👍

AtlasUI/AtlasUI/AtlasUIErrors.swift
```diff
@@ -52,12 +52,19 @@ extension UserPresentable {
 
 extension AtlasAPIError: UserPresentable {
 
+    func shouldDisplayGeneralMessage() -> Bool {
+        switch self {
+        case let .nsURLError(_, details): return details == nil
```
bonkey: Is it reasonable? Should details of NSURLError always go to the user? They're usually pretty reasonable, but I would double check it...

bonkey:
👍

AtlasUI/AtlasUI/AtlasUIErrors.swift
hani-ibrahim: I think yes as they already have localized description to be displayed to the end user. Any apple app that have a connection problem display these errors.

However, they are pretty many errors 😖 
https://developer.apple.com/reference/foundation/1508628-url_loading_system_error_codes

bonkey:
:+

Issue #335, opened 2016-09-27
Error handling: Payment method is unavailable


bonkey:
**Given** User has selected payment method
**When** Risk check is triggered (e.g. due to a changed address)
**And** Disallowed selected payment method
**Then** Banner is presented with information to select another payment method

Needs to #327

Issue #334, opened 2016-09-27
Error handling: Article is out of stock


bonkey:
**Given** User is in the checkout process
**When** the selected item gets out of stock.
**Then** "Oops" message is shown: Sorry, someone else bought this article faster than you. Please choose another one.

Needs #328
Related to #333

hani-ibrahim:
- need to check if the `updateCheckout` API will return out of stock or not

Issue #333, opened 2016-09-27
Error handling: Price changed just before order


bonkey:
**Given** User has completed all checkout data and order is possible
**When** User press "Place order"
**And** Price has changed
**Then** Order **shouldn't be placed**, and the user should be informed that price has changed.

Related to #332

hani-ibrahim:
- need to find a way to test it

Issue #332, opened 2016-09-27
Error handling: Price changed during checkout


bonkey:
**Given** User is in any of the steps of checkout process
**When** User gets information about the article price change in cart
**Then** Banner telling about the change (Headline: price changed // Error text: The price is now: Y) should be presented.

_Note: Specific case and most common case for different size prices is already handled in #298_

PR #396, opened 2016-10-04
Erros Handling

hani-ibrahim:
working on #329 - #331 - #332 - #333 - #334 - #335 

- Adding reachability to display the banner automatically
- Slide animating the banner instead of fading
- Displaying error if no address is selected
- Displaying banner if the price is changed
- Displaying error if no the payment method is no longer available (assuming the checkout API will not return the selected payment)
- Recreating the checkout object right before placing the order to double check the availability and the price


hani-ibrahim:
👍

AtlasUI/AtlasUI/AtlasUIViewController.swift
```diff
@@ -17,7 +17,6 @@ public class AtlasUIViewController: UIViewController {
         mainNavigationController = UINavigationController(rootViewController: sizeSelectionViewController)
 
         super.init(nibName: nil, bundle: nil)
-        setupReachability()
```
bonkey: Aha! I told you, don't do too much in `init`!

hani-ibrahim: I already fixed it :D

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
...
+        #if (arch(i386) || arch(x86_64)) && os(iOS)
+            return false
+        #else
+            return true
+        #endif
+    }()
+
+    private var notifierRunning = false
+    private var reachabilityRef: SCNetworkReachability?
+    private let reachabilitySerialQueue = dispatch_queue_create("uk.co.ashleymills.reachability", DISPATCH_QUEUE_SERIAL)
```
bonkey: Wouldn't be good to have own queue id?

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
...
+public enum ReachabilityError: ErrorType {
+    case FailedToCreateWithAddress(sockaddr_in)
+    case FailedToCreateWithHostname(String)
+    case UnableToSetCallback
+    case UnableToSetDispatchQueue
+}
+
+public let ReachabilityChangedNotification = "ReachabilityChangedNotification"
+
+func callback(reachability:SCNetworkReachability, flags: SCNetworkReachabilityFlags, info: UnsafeMutablePointer<Void>) {
```
bonkey: private?

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+
+public class Reachability: NSObject {
+
+    public typealias NetworkReachable = (Reachability) -> ()
+    public typealias NetworkUnreachable = (Reachability) -> ()
+
+    public enum NetworkStatus: CustomStringConvertible {
```
bonkey: All public -> internal.

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
...
+func callback(reachability:SCNetworkReachability, flags: SCNetworkReachabilityFlags, info: UnsafeMutablePointer<Void>) {
+    let reachability = Unmanaged<Reachability>.fromOpaque(COpaquePointer(info)).takeUnretainedValue()
+
+    dispatch_async(dispatch_get_main_queue()) {
+        reachability.reachabilityChanged(flags)
+    }
+}
+
+
+public class Reachability: NSObject {
```
bonkey: I'd love to have own code for it... I know it's impossible now or maybe even ever.

AtlasUI/AtlasUI/BannerErrorViewController.swift
```diff
@@ -60,24 +60,45 @@ class BannerErrorViewController: UIViewController {
...
+        view.alpha = 1
+
+        UIView.animateWithDuration(0.5) { [weak self] in
+            self?.containerView.transform = CGAffineTransformIdentity
+        }
+    }
+
+    func hideBanner() {
+        let bannerHeight = containerView.bounds.height
+        UIView.animateWithDuration(0.5, animations: { [weak self] in
```
bonkey: All magic numbers like `0.5` should be in common consts used in all animations, named like slow, fast, etc.

AtlasUI/AtlasUI/BannerErrorViewController.swift
```diff
@@ -60,24 +60,45 @@ class BannerErrorViewController: UIViewController {
...
+    override func viewDidAppear(animated: Bool) {
+        super.viewDidAppear(animated)
+        showBanner()
+    }
+
+    func showBanner() {
+        containerView.transform = CGAffineTransformMakeTranslation(0, -containerView.bounds.height)
+        view.alpha = 1
+
+        UIView.animateWithDuration(0.5) { [weak self] in
```
bonkey: Meeep! Magic number detected (comment below)

AtlasUI/AtlasUI/AtlasUIViewController.swift
```diff
@@ -19,10 +23,52 @@ public class AtlasUIViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        reachability?.stopNotifier()
+        reachability = nil
```
bonkey: Isn't `nil` is enough and don't need to stop it? Then, whole `deinit` is unneeded.

AtlasUI/AtlasUI/AtlasUIViewController.swift
```diff
@@ -3,10 +3,14 @@
...
 
 import UIKit
+import AtlasSDK
 
 public class AtlasUIViewController: UIViewController {
 
     let mainNavigationController: UINavigationController
+    internal let bannerErrorViewController = BannerErrorViewController()
+    internal let fullScreenErrorViewController = FullScreenErrorViewController()
+    private var reachability: Reachability?
```
bonkey: Maybe better to keep it as separated or in `AtlasCheckout`?

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
+// swiftlint:disable variable_name
+// swiftlint:disable colon
+// swiftlint:disable type_body_length
+// swiftlint:disable line_length
+// swiftlint:disable trailing_whitespace
+// swiftlint:disable force_unwrapping
```
bonkey: BTW. probably it doesn't have tests, so please [exclude it from codecov](http://docs.codecov.io/docs/ignoring-paths)

AtlasUI/AtlasUI/Utils/Reachability.swift
```diff
@@ -0,0 +1,372 @@
...
+import Foundation
+
+public enum ReachabilityError: ErrorType {
+    case FailedToCreateWithAddress(sockaddr_in)
+    case FailedToCreateWithHostname(String)
+    case UnableToSetCallback
+    case UnableToSetDispatchQueue
+}
+
+public let ReachabilityChangedNotification = "ReachabilityChangedNotification"
```
bonkey: A different name for this, like "AtlasSDK." prefix or so.

AtlasUI/AtlasUI/Utils/Reachability.swift
bonkey: I know what it is. This _fancy_ code exists since the beginning of the iPhoneOD era.

And that's why I'd love to have own code for this, because:

1. First – would rather not to include anything 3rd party
2. Second – we don't need full blown reachability feature set

bonkey: This looks much lighter: http://stackoverflow.com/a/39783037/148459

hani-ibrahim:
👍

hani-ibrahim:
Needs https://github.com/zalando-incubator/atlas-ios/pull/450 first

hani-ibrahim:
@bonkey just finished the errors and sorry for this unreadable PR 😂

hani-ibrahim:
👍

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Errors.swift
```diff
@@ -27,19 +29,23 @@ public enum AtlasConfigurationError: AtlasErrorType {
...
-    case checkoutFailed(addresses: [UserAddress]?, cartId: String?, error: ErrorType)
+    case checkoutFailed(addresses: [UserAddress]?, cart: Cart?, error: ErrorType)
 
 }
 
 public enum AtlasCatalogError: AtlasErrorType {
 
     case outOfStock
+    case missingAddress
+    case priceChanged(newPrice: NSDecimalNumber)
```
bonkey: Maybe a typealias for `NSDecimalNumber` for such cases?

AtlasSDK/AtlasSDK/Models/CheckoutWithCart.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public struct CheckoutWithCart {
```
bonkey: Seems like naming could be better, more semantic, or in other words, how would you like to go into such naming (just one step lower)?

`StringWithStringWithStringWithDeliveryWithPaymentWithAddressWithAddressAndWithStringWithStringWithCartItemsWithStringsWithDeliveryWithMoneyWithMoney`

AtlasSDK/AtlasSDK/Utils/String.swift
```diff
@@ -17,4 +17,13 @@ extension String {
         return string.stringByReplacingOccurrencesOfString("  ", withString: "")
     }
 
+    var range: NSRange {
+        return NSRange(location: 0, length: characters.count)
+    }
+
+    var bracketsFree: String {
+        let regex = try? NSRegularExpression(pattern: "\\([^\\)]*\\)", options: .CaseInsensitive)
```
bonkey: Let's have a rule to explicitly name all the regex patterns (either of `NSRegularExpression` or even better `String` in separated `let`), to understand what's supposed to be there...

bonkey: And BTW. why do you want to remove anything in brackets, and have it as SDK-wide feature?

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -8,7 +8,14 @@ import AtlasSDK
 class CheckoutSummaryViewController: UIViewController, CheckoutProviderType {
 
     internal var checkout: AtlasCheckout
-    internal var checkoutViewModel: CheckoutViewModel
+    internal var checkoutViewModel: CheckoutViewModel {
+        didSet {
```
bonkey: Don't build any extended `didSet` without reasonable unit tests, please. It's a shortcut to troubles.

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -53,6 +60,40 @@ class CheckoutSummaryViewController: UIViewController, CheckoutProviderType {
         self.navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .Plain, target: nil, action: nil)
         self.navigationController?.navigationBar.accessibilityIdentifier = "checkout-summary-navigation-bar"
     }
+}
+
+extension CheckoutSummaryViewController {
```
bonkey: Hot stuff here in this extension...

hot

AtlasUI/AtlasUI/CheckoutViewModel.swift
```diff
@@ -93,7 +92,7 @@ extension CheckoutViewModel {
         if customer == nil {
             return .NotLoggedIn
         }
-        return (checkout?.payment.selected?.method == nil) ? .CheckoutIncomplete : .LoggedIn
+        return (checkout?.payment.selected?.method == nil) ? .CheckoutIncomplete : .CheckoutReady
```
bonkey: How about nice if-elseif-else here for all 3 cases?

AtlasUI/AtlasUI/Utils/AtlasReachability.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    internal func setupReachability() {
+        let viewController: AtlasUIViewController? = try? Atlas.provide()
+        guard let atlasUIViewController = viewController else { return }
+
+        do {
+            reachability = try AtlasUI_Reachability.reachabilityForInternetConnection()
+            try reachability?.startNotifier()
+        } catch {
+            AtlasLogger.logError("Reachability Configuration error")
```
bonkey:
```
} catch let error {
  AtlasLogger.logError(error)
}
```

AtlasSDK/AtlasSDK/Models/CheckoutWithCart.swift
hani-ibrahim: :D 
Yeah I don't like either. but how can I renamed it? 
I made it first as `CheckoutCart` but seams to be not understandable. Actually I could replace this whole thing with tuple if this is good

AtlasSDK/AtlasSDK/Utils/String.swift
hani-ibrahim: I'm not confident about putting it in the SDK too... probably internal? 
I used it only for `localizedTitleKey` and `localizedMessageKey`
`var localizedTitleKey: String { return "\(self.dynamicType).title.\(self)".bracketsFree }`

For the price case it's `case priceChanged(newPrice: NSDecimalNumber)` so  returning `"\(self.dynamicType).title.\(self)"` for this case would be `priceChanged(200)` which will not be found in the String files. So I'm trimming the associated value part

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
hani-ibrahim: that why I grouped them in one extension. 🔫 
- `injectCustomer` is really important to be modified at the didSet. because we always forgot and then find it out as a bug. This is happened many time with me, Ahmed and Gleb
- `checkPriceChange` - `checkPaymentMethod` these are the errors that depends on the previous value of the viewModel to check if they are changed and then errors would happens ... where could we put them?

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
bonkey: I have very bad memory, but AFAIR it never come without pain in my code to chain actions this way.

Just in this case you already had to have 7 conditions to check and imagine something will change around them.

Very tight coupling

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
bonkey: Might be better with didSet, but still, pretty complex rules fired implicitely when something touches it.

jump

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
hani-ibrahim: well, if thing is changed then it will be updated in one place, better than going through all the code where we modify the view model and recheck everything.

Check these 4 lines 
https://github.com/zalando-incubator/atlas-ios/pull/396/files#diff-46a836102c2c0505453fc4a19421829eL178

especially the first one. we always forgot to inject it. And these (copy & paste) 4 lines are now replaced with one simple line. 

I understand your point. But I'm not sure if it's a bad approach or not

bonkey: In this situation in the code I don't have a silver bullet. That's why I would just fasten seatbelts with tests and run.

hani-ibrahim:
👍

hani-ibrahim:
finally added some tests
testing view controller is very tricky 😖

bonkey:
@hani-ibrahim it's on my reading list... http://qualitycoding.org/testing-view-controllers/

AtlasUI/AtlasUITests/CheckoutSummaryViewControllerSpec.swift
```diff
@@ -0,0 +1,126 @@
...
+                    expect(viewController.checkoutViewModel.customer).toNot(beNil())
+                    expect(viewController.viewState).to(equal(CheckoutViewState.CheckoutReady))
+                    expect(atlasUIViewController.errorDisplayed).to(equal(false))
+
+                    // Remove Checkout
+                    self.clearState(viewController, atlasUIViewController: atlasUIViewController)
+                    viewController.checkoutViewModel.checkout = nil
+                    expect(viewController.checkoutViewModel.customer).toNot(beNil())
+                    expect(viewController.viewState).to(equal(CheckoutViewState.CheckoutIncomplete))
+                    expect(atlasUIViewController.errorDisplayed).to(equal(true))
```
bonkey: I'm pretty surprised, why these comment-prefixed blocks are not separated tests?

AtlasUI/AtlasUITests/CheckoutSummaryViewControllerSpec.swift
```diff
@@ -0,0 +1,126 @@
...
+                    done()
+                }
+            }
+
+            try! AtlasMockAPI.stopServer() // swiftlint:disable:this force_try
+        }
+    }
+
+
+    private func viewController(completion: ((CheckoutSummaryViewController, AtlasUIViewController) -> Void)) {
```
bonkey: `createCheckoutSummaryViewController` maybe or sth?

AtlasUI/AtlasUITests/CheckoutSummaryViewControllerSpec.swift
```diff
@@ -0,0 +1,126 @@
...
+        let salesChannelId = "82fe2e7f-8c4f-4aa1-9019-b6bde5594456"
+        let clientId = "CLIENT_ID"
+        let options = Options(clientId: clientId,
+                              salesChannel: salesChannelId,
+                              interfaceLanguage: interfaceLanguage,
+                              configurationURL: configURL)
+
+
+        AtlasCheckout.configure(options) { result in
+            guard let checkout = result.process() else { return XCTFail() }
```
bonkey: Maybe in direction of

```
expect(checkout).toNot(beNil(), description: "Kabooom!")
```

or 

```
expect { let checkout = result.process() /* ... */ }.toNot(throwError())
```

(I know, it's no for optional unboxing and we don't test this part, but maybe it could be clearer?

bonkey:
👍

Issue #423, opened 2016-10-05
Change "Error" header for "Out of stock" full screen


bonkey:
### Message from Atlas Testing sent on 2016-10-05 15:24:12 UTC###

> Error is not really a good headline for out of stock message.
> 
> -------------
> iPhone 6
> iPhone OS 9.3.3

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/398497/feedback/632881](https://rink.hockeyapp.net/manage/apps/398497/feedback/632881)

bonkey:
![size-error](https://cloud.githubusercontent.com/assets/39100/19150278/3527fa78-8bc6-11e6-946b-7341dff177d7.jpg)

PR #460, opened 2016-10-06
#423: rename error header

hani-ibrahim:
renaming error header for out of stock screen (found in Bug bashing) #423


hani-ibrahim:
👍

ahmed-shehata:
👍

PR #462, opened 2016-10-07
#456: Fixed loading double loginVC

ahmed-shehata:
#456: Fixed loading double loginVC


ahmed-shehata:
👍

AtlasUI/AtlasUI/OAuth2AuthorizationHandler.swift
```diff
@@ -21,9 +21,11 @@ struct OAuth2AuthorizationHandler: AuthorizationHandler {
         let loginViewController = OAuth2LoginViewController(loginURL: self.loginURL, completion: completion)
         let navigationController = UINavigationController(rootViewController: loginViewController)
 
-        Async.main {
-            navigationController.modalPresentationStyle = .OverCurrentContext
-            topViewController.presentViewController(navigationController, animated: true, completion: nil)
+        if !topViewController.isKindOfClass(OAuth2LoginViewController.classForCoder()) {
```
hani-ibrahim: I would not fix it that way. Still if the user was fast enough, I think he can open it multiple time.

I would use this library to fix these kind of problems
https://github.com/soffes/RateLimit

ahmed-shehata: Not really, once you call presentViewController, the topViewController would report "LoginViewController". Even if the animation transition did not complete.

hani-ibrahim: `presentViewController` is in `Async.main` which could give a delay and a chance for the user to retry. 
We have this problem in many places as creating an address. and these problem should be fixed in a proper way not a hacky way :P

ahmed-shehata: It is not a "hacky" way, but rather checks wherever in the code the "authorize" function was called and then checks to see if the loginVC is already on top. All of this runs in the main thread so there should not be any race condition.

hani-ibrahim: @bonkey your input :-)

AtlasUI/AtlasUI/OAuth2AuthorizationHandler.swift
```diff
@@ -22,8 +22,10 @@ struct OAuth2AuthorizationHandler: AuthorizationHandler {
         let navigationController = UINavigationController(rootViewController: loginViewController)
 
         Async.main {
-            navigationController.modalPresentationStyle = .OverCurrentContext
-            topViewController.presentViewController(navigationController, animated: true, completion: nil)
+            if !topViewController.isKindOfClass(OAuth2LoginViewController.classForCoder()) {
```
bonkey: Could be rather `if !(topViewController is OAuth2LoginViewController)`, but it's not the case here. I would just close this PR and talk different approaches.

bonkey:
So, closing, sir.

PR #469, opened 2016-10-11
#issue: 421 - saving multiple address during connection problem

hani-ibrahim:
I tried to use `RateLimit` library ... but unfortunately I'm not satisfied with the result.

If I set the rateLimit for the submit button to be 1 sec then If I pressed on the submit button 10 times in 5 sec where the times between each click is 0.5 sec... then the submit button action will be executed only one time. 
However for our case. the first case is not successful because of internet connection so clicking 10 times in 5 seconds will not do anything even if the internet is back ... the user must stop pressing for 1 second and try again.

I'll try to find another approach


bonkey:
So, shall we reject this PR? Maybe we should sit over this issue with brain storm or so?

hani-ibrahim:
I solved it with enabling and disabling the save button. But my problem with this is that I'm tracking all the failures to re-enable the button.

Not so confident with that

hani-ibrahim:
👍

bonkey:
👍

bonkey:
BTW. travis passed 7 minutes ago, maybe we should end in few minutes, meaning 10 minutes later. Shall we change requirements on statuses? :)

hani-ibrahim:
yeah we can try to make travis the required one instead of Buddy build and see

PR #482, opened 2016-10-14
#481: Dropped unused Unit model

bonkey:
#481: Dropped unused Unit model


bonkey:
It was poking my eye for long but dug deeper when size/availability issue was raised and seems not used at all. I hope I didn't break anything and it's just leftover :)

hani-ibrahim:
👍

Issue #491, opened 2016-10-17
Ray-Ban loads without loader displayed.


hani-ibrahim:
Note: It is one size

Issue #411, opened 2016-10-05
Screen ends up with error message + infinite loading only


bonkey:
### Message from Atlas Testing sent on 2016-10-05 13:55:00 UTC###

> No idea what this is
> 
> 
> -------------
> iPhone 6
> iPhone OS 9.3.1

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/398497/feedback/632819](https://rink.hockeyapp.net/manage/apps/398497/feedback/632819)

ahmed-shehata:
1 - Find out what is the exact scenario (network, blocking, etc.. ?)
2 - Handle if special case (in this case blocking error OR handle refresh somehow)

bonkey:
![spinner-error-only](https://cloud.githubusercontent.com/assets/39100/19150134/90cb12ee-8bc5-11e6-966b-eab3d979ab07.jpg)

Issue #142, opened 2016-08-17
Move handling of single unit out of SizeSelectionViewController


bonkey:
Handle single unit case inside `AtlasCheckout.presentCheckoutView` and present either `SizeSelectionViewController` or `CheckoutSummaryViewController` instead of always opening `SizeSelectionViewController` and skipping the screen if there's one unit.

It would help with some "hacks" in those controlleors

Could be related to #138

PR #492, opened 2016-10-17
#411: - Size Selection Refactor

hani-ibrahim:
- #411: fixing errors occurred on the size list screen that if the view is empty then the error would be full screen error
- #491: fixing the problem where the loader may not appear
- #142: refactor size list selection screens


hani-ibrahim:
👍

AtlasUI/AtlasUI/SizeListSelectionViewController.swift
```diff
@@ -7,15 +7,22 @@ import AtlasSDK
 
 final class SizeListSelectionViewController: UIViewController, CheckoutProviderType {
 
-    private let article: Article
     internal let checkout: AtlasCheckout
+    private let sku: String
+    private var tableViewDelegate: SizeListTableViewDelegate? {
+        didSet {
+            tableView.dataSource = tableViewDelegate
+            tableView.delegate = tableViewDelegate
```
bonkey: Wouldn't you want to have separated classes for dataSource and delegate?

AtlasUI/AtlasUI/SizeListSelectionViewController.swift
```diff
@@ -69,56 +73,44 @@ extension SizeListSelectionViewController: UIBuilder {
...
-            self.loaderView.hide()
-
-            guard let customer = result.process() else { return }
-            let selectedArticleUnit = SelectedArticleUnit(article: self.article, selectedUnitIndex: indexPath.row)
-
-            self.loaderView.show()
-            self.checkout.createCheckoutViewModel(forArticleUnit: selectedArticleUnit) { result in
-                self.loaderView.hide()
+        checkout.client.customer { [weak self] result in
+            guard let customer = result.process(forceFullScreenError: !userSelected) else {
```
bonkey: What does it mean that `userSelected`? Feels like better name is needed in this context.

AtlasUI/AtlasUI/SizeListSelectionViewController.swift
```diff
@@ -69,56 +73,44 @@ extension SizeListSelectionViewController: UIBuilder {
...
+                self?.displayCheckoutSummaryViewController(checkoutViewModel, userSelected: userSelected)
             }
         }
     }
 
-    private func displayCheckoutSummaryViewController(checkoutViewModel: CheckoutViewModel) {
+    private func displayCheckoutSummaryViewController(checkoutViewModel: CheckoutViewModel, userSelected: Bool) {
         let checkoutSummaryVC = CheckoutSummaryViewController(checkout: checkout, checkoutViewModel: checkoutViewModel)
-        self.showViewController(checkoutSummaryVC, sender: self)
+        navigationController?.pushViewController(checkoutSummaryVC, animated: userSelected)
```
bonkey: Here as well. Seems pretty vague that selected user makes something animated or not. I can imagine the case, but would be good to understand it from the name of the variable without imagination.

AtlasUI/AtlasUI/Utils/UserMessage.swift
```diff
@@ -35,12 +35,17 @@ struct UserMessage {
         bannerErrorViewController.hideBanner()
     }
 
-    static func displayError(error: ErrorType) {
+    static func displayError(error: ErrorType, forceFullScreenError fullScreen: Bool = false) {
```
bonkey: Would be better to use a separated method instead of this bool argument.

hani-ibrahim: @bonkey Done but I don't like it now
It's more confusing for the reader of `UserMessage` class than a boolean that is clear what it is doing

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/ConfigClient.swift
```diff
@@ -25,12 +25,19 @@ struct ConfigClient: Configurator {
...
                 guard let json = response.body, config = Config(json: json, options: self.options) else {
-                    return completion(.failure(AtlasConfigurationError.incorrectConfigServiceResponse))
+                    dispatch_async(dispatch_get_main_queue()) {
+                        completion(.failure(AtlasConfigurationError.incorrectConfigServiceResponse))
+                    }
+                    return
+                }
+                dispatch_async(dispatch_get_main_queue()) {
+                    completion(.success(config))
                 }
```
bonkey: Oh, better don't go to main thread in SDK, please. Can't it be handled in AtlasUI only?

AtlasUI/AtlasUI/Utils/UserMessage.swift
bonkey: Maybe it's just from your chair? From code review perspective, it is noticeably better. Logic is close to the argument and it's not passed from far, far away land.

bonkey:
👍

AtlasSDK/AtlasSDK/ConfigClient.swift
hani-ibrahim: we agreed before to use the main thread here 
https://github.com/zalando-incubator/atlas-ios/blob/master/AtlasSDK/AtlasSDK/APIClient.swift#L56
but seams that config API was missed out and I got a crash because of it

PR #496, opened 2016-10-18
Dropping semi-sensitive information from source

bonkey:
Tiny clean-ups for #267


bonkey:
👍

AtlasMockAPI/AtlasMockAPI/JSON/POST*|checkouts.json
```diff
@@ -39,6 +39,6 @@
             },
             "external_payment" : false
         },
-        "selection_page_url" : "https://payment-gateway.kohle-integration.zalan.do/payment-method-selection-session/595f8ebf-8525-4f66-91cf-5fb3b9ad86c1/selection"
+        "selection_page_url" : "https://localhost.charlesproxy.com:9080/payment-method/595f8ebf-8525-4f66-91cf-5fb3b9ffffc1"
```
hani-ibrahim: does that means I'll need to install charlesproxy ?

bonkey: Nope.

1. It doesn't change anything: doesn't work similar way like it didn't before this change
2. `localhost.charlesproxy.com` is just a DNS record, which points to `127.0.0.1`

AtlasDemo/AtlasDemo.xcodeproj/project.pbxproj
```diff
@@ -7,53 +7,53 @@
...
-		44EBCE2FE12DE8193EE4F38738F84B2E /* ImageIO.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 1E9AC22E8436A9E4B159173A8D00CE86 /* ImageIO.framework */; };
-		4C7F901DEBA1E479432682A6E43A6266 /* VideoToolbox.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B5D445C1DAAC347B64D5AFED9B5B140B /* VideoToolbox.framework */; };
-		5354EC81B56390B95BAA15D4F3965B41 /* CoreTelephony.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E4F833A4CE21235444066F2613EFB09A /* CoreTelephony.framework */; };
-		5C38BE2E6C30F953FDD8E2C0891B4EFF /* CoreAudio.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A09363B8394CF5F3F4FBFE3E3AB2F5B1 /* CoreAudio.framework */; };
-		5DCE7A0D970BD405625229A88DAA3D6F /* AtlasSDK.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FD1554ECC3021D4DC35E014009832E9C /* AtlasSDK.framework */; };
+		E283E83F8AB0B137D725B276F2A134E7 /* Alamofire.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 4D539715BD2C9531ED72B24D9B4A98FF /* Alamofire.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
bonkey: Just result of `./calypso.rb clean xunique`

hani-ibrahim: Aha, Now I got why every couple of weeks my tree is missed up!
I like to order files in a meaningful order (not alphabetically)
see the difference between my tree and calypso tree 😕 😖 
screen shot 2016-10-18 at 3 52 03 pm
screen shot 2016-10-18 at 3 51 10 pm

hani-ibrahim: - what conflicts could happens? I don't see any problem if the files are not ordered alphabetically. Actually I got many conflicts because of this script as it change the whole tree and if I added a new file then I will get a conflicts. 
- creating subgroups will not fix the problem as it will be ordered alphabetically.

For my opinion, automation is not always good specially if it adds zero value and introduce problems.

I'm not saying removing the `./calypso.rb clean` completely, but the part where it modify the tree is not needed

bonkey: BTW. they suggest to unique on post-build... Might be interesting idea.

https://github.com/truebit/xUnique#add-xunique-to-xcode-post-action

bonkey: But my conflict was easily merged automatically.

Maybe going with newer Xcode project format would be also an alternative?

hani-ibrahim: what do you mean by newer?

hani-ibrahim:
👍

AtlasDemo/AtlasDemo.xcodeproj/project.pbxproj
bonkey: We use "Xcode 3.2-compatible", and I thought next one (6.3) bu default uses XML, but it's seems separated things: structure and format. And moreover XML format seems even more flaky: https://github.com/CocoaPods/CocoaPods/issues/1275

hani-ibrahim: yeah XML like storyboard is not good

bonkey:
👍

PR #499, opened 2016-10-19
#486: Unify the loader

hani-ibrahim:
Working on #486

### DIFF
--- a/AtlasUI/AtlasUI/LoaderView.swift
+++ b/AtlasUI/AtlasUI/LoaderView.swift
@@ -3,6 +3,7 @@
 //
 
 import UIKit
+import AtlasSDK
 
 class LoaderView: UIView {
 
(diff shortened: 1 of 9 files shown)


hani-ibrahim:
👍

AtlasUI/AtlasUI/LoaderView.swift
```diff
@@ -27,6 +28,14 @@ class LoaderView: UIView {
         hidden = true
     }
 
+    internal static func displayLoader(block: (() -> Void) -> Void) {
```
bonkey: ihwcaj6r061wc

bonkey: I get the reason, but it will kick back one day...

bonkey:
👍

AtlasUI/AtlasUI/LoaderView.swift
hani-ibrahim: what about moving it to usermessage?

PR #516, opened 2016-10-27
#513: SKUs used in demo app

hani-ibrahim:
Working on #513 

- First there were more than 20 SKUs and the API return only 20 unsorted SKUs per page so that's why sometimes I get no article in stock
- Second removed many SKUs that are out of stock ... left only few that are out of stock for testing


hani-ibrahim:
👍

bonkey:
👍

Issue #498, opened 2016-10-19
Create view controller to present mobile-friendly legal information


bonkey:
Allows to handle problems with external links from:
- payment #444
- ToC webviews #403 
- login
- summary view

Because the solution is based on data provider which is about to expire, as a fallback there will be a regular ToC page.

PR #521, opened 2016-10-28
#498: LegalController with fallback to ToC from config service

bonkey:
# 498: LegalController with fallback to ToC from config service


bonkey:
👍

bonkey:
👍

AtlasDemo/AtlasDemo.xcodeproj/xcshareddata/xcschemes/AtlasDemo.xcscheme
```diff
@@ -109,7 +109,7 @@
       <CommandLineArguments>
          <CommandLineArgument
             argument = "USE_MOCK_API"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
hani-ibrahim: :triumph:

AtlasUI/AtlasUI/PaymentSelectionViewController.swift
```diff
@@ -83,18 +83,27 @@ final class PaymentViewController: UIViewController, UIWebViewDelegate {
...
+        if !error.isWebKitError {
+            UserMessage.displayError(error)
+        }
+    }
+
+}
+
+private extension NSError {
+
+    var isWebKitError: Bool {
```
hani-ibrahim: not sure about `WebKitError` because it's not clear what is the purpose of it or why we are checking on it. 
The problem is when I return `false` from `shouldStartLoadWithRequest` it return this `WebKitErrorDomain` but is't not actual an error so I don't want to display an error to the user

hani-ibrahim: yeah true, but then someone could remove it in the future because he will not understand why it's done and he may not figure out how to trigger it

bonkey: Well, I was close to remove it instead of renaming, because it was unclear what it does 👹

hani-ibrahim:
👍

bonkey:
👍

bonkey:
:+1:

hani-ibrahim:
👍

PR #527, opened 2016-10-31
#456: Tapping on login button twice

hani-ibrahim:
working on #456

### DIFF
--- a/AtlasDemo/AtlasDemo/ProfileViewController.swift
+++ b/AtlasDemo/AtlasDemo/ProfileViewController.swift
@@ -65,8 +65,8 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
-                guard let userError = error as? AtlasUserError where userError != AtlasUserError.userCancelled else {
-                    print("SWALLOWED: ", error)
+                guard (error as? AtlasUserError) != AtlasUserError.userCancelled else {
+                    AtlasLogger.logError(error)
                     return
                 }
 
--- a/AtlasDemo/AtlasDemo/ProfileViewController.swift
+++ b/AtlasDemo/AtlasDemo/ProfileViewController.swift
@@ -65,6 +65,11 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
+                guard let userError = error as? AtlasUserError where userError != AtlasUserError.userCancelled else {
+                    print("SWALLOWED: ", error)
+                    return
+                }
+
                 self.displayError(error)
 
             case .success(let customer):
--- a/AtlasDemo/AtlasDemo/ProfileViewController.swift
+++ b/AtlasDemo/AtlasDemo/ProfileViewController.swift
@@ -65,12 +65,7 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
-                switch error {
-                case LoginError.userCancelled:
-                    break
-                default:
-                    self.displayError(error)
-                }
+                self.displayError(error)
 
             case .success(let customer):
                 self.avatar.image = UIImage(named: "user")
--- a/AtlasDemo/AtlasDemo/ProfileViewController.swift
+++ b/AtlasDemo/AtlasDemo/ProfileViewController.swift
@@ -65,7 +65,12 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
-                self.displayError(error)
+                switch error {
+                case LoginError.userCancelled:
+                    break
+                default:
+                    self.displayError(error)
+                }
 
             case .success(let customer):
                 self.avatar.image = UIImage(named: "user")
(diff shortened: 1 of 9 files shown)


hani-ibrahim:
👍

AtlasDemo/AtlasDemo/ProfileViewController.swift
```diff
@@ -65,7 +65,12 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
-                self.displayError(error)
+                switch error {
```
bonkey: Maybe guard to make it more obvious?

```
    case let .success(result):
        print("SUCCESS:", result)
    case let .failure(error):
        guard let loginError = error as? LoginError, loginError != LoginError.userCancelled else {
            print("SWALLOWED:", error)
            return
        }
        print("FAILURE:", error)
    }
```

AtlasUI/AtlasUI/CheckoutSummaryActionsHandler.swift
```diff
@@ -64,10 +64,28 @@ extension CheckoutSummaryActionsHandler {
...
+                guard let customer = self.userLoginResultProcess(result) else { return }
+                self.generateCheckout(customer)
+            }
+        }
+    }
+
+    private func userLoginResultProcess(result: AtlasResult<Customer>) -> Customer? {
+        switch result {
+        case .failure(let error):
+            switch error {
```
bonkey: `guard` as above?

AtlasDemo/AtlasDemo/ProfileViewController.swift
bonkey: And BTW. do we need to handle it here? Two other places are fine. Here I would just rebuild the flow a bit and in the end didn't care about the cancellation as it should not really be a burden of 3rd party devs

AtlasUI/AtlasUI/AtlasUIErrors.swift
```diff
@@ -92,3 +92,9 @@ extension AtlasCheckoutError: UserPresentable {
 extension LoginError: UserPresentable { }
 
 extension AtlasConfigurationError: UserPresentable { }
+
+public enum AtlasUserError: UserPresentable {
+
+    case userCancelled
```
bonkey: We should consider user's errors like heart stroke, too ;-)

AtlasDemo/AtlasDemo/ProfileViewController.swift
hani-ibrahim: but I think we need to handle it otherwise we will get an error if the user cancelled the login. I'm not use actually if we should provide the login in the profile. I think the login inside the SDK is enough

hani-ibrahim:
👍

AtlasDemo/AtlasDemo/ProfileViewController.swift
```diff
@@ -65,6 +65,11 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
+                guard let userError = error as? AtlasUserError where userError != AtlasUserError.userCancelled else {
+                    print("SWALLOWED: ", error)
```
bonkey: Really? PRINT? 👹 

copying

AtlasUI/AtlasUI/Utils/AtlasResult.swift
```diff
@@ -9,15 +9,15 @@ extension AtlasResult {
...
-            case AtlasUserError.userCancelled:
-                break
-            default:
-                if fullScreen {
-                    UserMessage.displayErrorFullScreen(error)
-                } else {
-                    UserMessage.displayError(error)
-                }
+            guard let userError = error as? AtlasUserError where userError != AtlasUserError.userCancelled else {
+                print("SWALLOWED: ", error)
```
bonkey: print

hani-ibrahim: I copied and paste from your comment 😂

hani-ibrahim:
👍

AtlasDemo/AtlasDemo/ProfileViewController.swift
```diff
@@ -65,8 +65,8 @@ class ProfileViewController: UIViewController {
         AppSetup.checkout?.client.customer { result in
             switch result {
             case .failure(let error):
-                guard let userError = error as? AtlasUserError where userError != AtlasUserError.userCancelled else {
-                    print("SWALLOWED: ", error)
+                guard (error as? AtlasUserError) != AtlasUserError.userCancelled else {
+                    AtlasLogger.logError(error)
```
bonkey: Do we need to log it?

bonkey:
👍

AtlasDemo/AtlasDemo/ProfileViewController.swift
hani-ibrahim: probably it will be used as a funnel with user tracking

bonkey: I like to bet when there's bigger chance to win than lose. Giving 5€ against.

premature

hani-ibrahim: I don't bet :P

Issue #547, opened 2016-11-09
Support "pod try" command with Example project


bonkey:

bonkey:
AtlasDemo is too depended on `AtlasUI` and `AtlasMockAPI` libraries, so it's too troublesome to support it.

PR #548, opened 2016-11-09
#547: Cleanups in Demo app

bonkey:
While preparing #547 I've found out some leftovers. While the issue won't be implemented, cleanups are needed to be integrated.


bonkey:
👍

hani-ibrahim:
👍

PR #559, opened 2016-11-15
#524: Removed custom sorting with SizeMap

bonkey:
#524: Removed custom sorting with SizeMap


AtlasDemo/AtlasDemo/DemoArticle.swift
```diff
@@ -254,8 +249,5 @@ public func == (lhs: DemoArticle.Unit, rhs: DemoArticle.Unit) -> Bool {
 }
 
 public func < (lhs: DemoArticle.Unit, rhs: DemoArticle.Unit) -> Bool {
```
hani-ibrahim: does this method actually needed?

hani-ibrahim:
👍

AtlasDemo/AtlasDemo/DemoArticle.swift
bonkey: When we remove sorting – no. So I removed.

bonkey:
👍

hani-ibrahim:
👍

PR #570, opened 2016-11-21
#567: Finished AtlasSDK migration to Swift 3.x

bonkey:
#567: Finished AtlasSDK migration to Swift 3.x


AtlasSDK/AtlasSDK/Endpoints/Endpoint.swift
```diff
@@ -35,16 +35,7 @@ extension Endpoint {
...
-            guard let parameters = parameters else { return nil }
-            do {
-                let data = try JSONSerialization.data(withJSONObject: parameters, options: [.prettyPrinted])
-                return String(data: data, encoding: String.Encoding.utf8)
-            } catch let error {
-                AtlasLogger.logError(error)
-                return nil
-            }
-        }() ?? "<NO PARAMETERS>"
+        let params = String(withJSONObject: parameters) ?? "<NO PARAMETERS>"
```
hani-ibrahim: Nice

AtlasSDK/AtlasSDK/Utils/String.swift
```diff
@@ -6,9 +6,19 @@ import Foundation
 
 extension String {
 
-    var whiteCharactersFreeString: String {
+    var newlinesRemoved: String {
```
hani-ibrahim: I'm not removing new lines here, you see that I'm converting `"   "` with `""` the reason for this is that the json returned from the backend has a lot of spaces in the json because it is formatted and this makes the console looks bad and crowded

bonkey: So, `whitespaceCompacted`?

PR #575, opened 2016-11-21
#571: First step in AtlasUI migration to Swift 3.x

bonkey:
At least it compiles


bonkey:
@hani-ibrahim I only have no idea, why again I got your changes in PR, when they should be already there...

bonkey:
Ah, because my push didn't come here on time ;-)

AtlasUI/AtlasUI/BannerErrorViewController.swift
```diff
@@ -96,24 +96,29 @@ extension BannerErrorViewController: UIBuilder {
...
 
     func configureConstraints() {
-        containerView.snapAnchorToSuperView(.top)
-        containerView.snapAnchorToSuperView(.right)
-        containerView.snapAnchorToSuperView(.left)
-        stackView.fillInSuperView()
-        cancelIconLabel.snapAnchorToSuperView(.top, constant: BannerErrorViewController.topMargin)
-        cancelIconLabel.snapAnchorToSuperView(.right, constant: -10)
-        cancelButton.fillInSuperView()
+        containerView.snap(anchor: self.view.topAnchor, toAnchor: self.view.superview?.topAnchor)
```
bonkey: @hani-ibrahim Here are some casualties. Feel free to adjust, but it's hard to get neat idea at 9 pm when you're in hurry to get something working :)

AtlasUI/AtlasUI/AddressFormStackView.swift
```diff
@@ -3,6 +3,29 @@
 //
 
 import UIKit
+// FIXME: comparison operators with optionals were removed from the Swift Standard Libary.
```
hani-ibrahim: huh?

AtlasUI/AtlasUI/BannerErrorViewController.swift
hani-ibrahim: but why you changed it? 
What I wanted by this is to hide the `self.view.*` and make it more readable. Now it not readable for me and hard to understand as (for example) what is the relation between the `cancelIconLabel` and `self.view`.

I would stick with the old one

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
```diff
@@ -16,22 +16,21 @@ class CheckoutSummaryViewController: UIViewController {
 
     var viewModel: CheckoutSummaryViewModel {
         didSet {
-            setupNavigationBar()
-            rootStackView.configureData(viewModel)
+            viewModelDidSet()
```
hani-ibrahim: I tried this before the problem here if the `viewModelDidSet` changed the `viewModel` again then it will enter a loop.

But the rule I found so far is:
- If you changed a property inside its didSet then the didSet will not be triggered again
- If you moved the didSet to another method and changed the property inside this method then the didSet will be triggered again resulting in an infinite loop

I'm not sure if this is modifying the viewModel inside or but it could be under some conditions as prices changes for example (not sure)
But for sure this is something that I would just avoid

hani-ibrahim: I tired now with Xcode:

This create an infinite loop
``` swift
class TEST {

    var x: Int? {
        didSet {
            changeX()
        }
    }

    func changeX() {
         x = 5
    }

}

let test = TEST()
test.x = 10
print(test.x as Any)
```


and this doesn't
``` swift
class TEST {

    var x: Int? {
        didSet {
            x = 5
        }
    }

}

let test = TEST()
test.x = 10
print(test.x as Any)
```

AtlasUI/AtlasUI/LoggedInAddressListActionHandler.swift
```diff
@@ -5,20 +5,20 @@
 import Foundation
 import AtlasSDK
 
-struct LoggedInAddressListActionHandler: AddressListActionHandler {
+class LoggedInAddressListActionHandler: AddressListActionHandler {
```
bonkey: @hani-ibrahim Another possibly tricky change. Did it "accidentally", when I changed `strategyCompletion` to a field. In this case `self` in [line 18](https://github.com/zalando-incubator/atlas-ios/pull/575/commits/ebacf87e67dbe143988b477997257f3f1b67609c#diff-7971a6dbc2cf2dd2b865f48d7834586fR18) was mutated, but it was only a copy. Didn't dig deeper, but seems like `@escaping` and structs needs deeper thoughts – not only in Swift 3, but with 3 it's guarded better by a compiler.

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
hani-ibrahim: Hmm, but still I'm not sure if this is actually the case or not and I don't remember when or where that happened with me 😕

AtlasUI/AtlasUI/Utils/Async.swift
```diff
@@ -6,71 +6,65 @@ import Foundation
...
 
-    static func background() -> dispatch_queue_t {
-        return dispatch_get_global_queue(QOS_CLASS_BACKGROUND, 0)
+    static func background() -> DispatchQueue {
+        return DispatchQueue.global(qos: DispatchQoS.QoSClass.background)
     }
 
 }
 
 struct Async {
```
hani-ibrahim: will you remove it?

AtlasUI/AtlasUI/AddressFormStackView.swift
bonkey: Xcode did it

AtlasUI/AtlasUI/Utils/Async.swift
bonkey: Async? Not sure yet.

AtlasUI/AtlasUI/BannerErrorViewController.swift
bonkey: Would love to keep it, but there're different `NSLayoutXAxisAnchor`, `NSLayoutYAxisAnchor` and they can't be returned as `NSLayoutAnchor<AnyObject>` from `anchor()`. Check out and play around. Would be great to have something similar back. However it's only 5 calls, so it's academic task so far.

AtlasUI/AtlasUI/CheckoutSummaryViewController.swift
bonkey: Haven't run tests or app, yet 👹

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -207,57 +207,55 @@ extension LoggedInSummaryActionHandler {
 
 extension LoggedInSummaryActionHandler {
 
-    fileprivate func updateDataModel(_ addresses: CheckoutAddresses?) {
+    fileprivate func updateDataModel(addresses: CheckoutAddresses?, usingCartCheckout: Bool = false) {
```
hani-ibrahim: I think by default it should use the `CartCheckout` as I need it to be nil in only one case which is "deleting the address"
may be rename it to `byRemovingCartCheckout: Bool = false`

AtlasUI/AtlasUI/LoggedInAddressListActionHandler.swift
hani-ibrahim: actually that's why I didn't create a property and created a function to set the completion inside, as there is no need to make it class and then need to handle `weak self` everywhere 🔫

hani-ibrahim: you still need to handle the `weak self` in the rest of the class. However I would still prefer it to be struct since it's very simple

AtlasUI/AtlasUI/Utils/UIView.swift
```diff
@@ -4,31 +4,14 @@
 
 import Foundation
 
-enum ViewAnchor {
```
hani-ibrahim: now I see why

I ended up doing something like this

``` swift
enum ViewAnchor {
    case top
    case right
    case bottom
    case left

    private func xAnchorForView(view: UIView) -> NSLayoutXAxisAnchor? {
        switch self {
        case .top: return nil
        case .right: return view.rightAnchor
        case .bottom: return nil
        case .left: return view.leftAnchor
        }
    }

    private func yAnchorForView(view: UIView) -> NSLayoutYAxisAnchor? {
        switch self {
        case .top: return view.topAnchor
        case .right: return nil
        case .bottom: return view.bottomAnchor
        case .left: return nil
        }
    }

}


func snapAnchorToSuperView(anchor: ViewAnchor, constant: CGFloat = 0) {
        guard let superview = superview else { return }
        translatesAutoresizingMaskIntoConstraints = false
        var constraint: NSLayoutConstraint?
        if let viewXAnchor = anchor.xAnchorForView(view: self), let superViewXAnchor = anchor.xAnchorForView(view: superview) {
            constraint = viewXAnchor.constraint(equalTo: superViewXAnchor)
        } else if let viewYAnchor = anchor.yAnchorForView(view: self), let superViewYAnchor = anchor.yAnchorForView(view: superview) {
            constraint = viewYAnchor.constraint(equalTo: superViewYAnchor)
        }
        constraint?.constant = constant
        constraint?.isActive = true
    }
```


What do you think?

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -204,57 +207,55 @@ extension LoggedInSummaryActionHandler {
 
 extension LoggedInSummaryActionHandler {
 
-    private func updateDataModel(addresses: CheckoutAddresses?) {
+    fileprivate func updateDataModel(addresses: CheckoutAddresses?, usingCartCheckout: Bool = false) {
```
hani-ibrahim: I don't know why the comment is hidden but here is it again in case you didn't see it

> I think by default it should use the `CartCheckout` as I need it to be nil in only one case which is "deleting the address"
may be rename it to `byRemovingCartCheckout: Bool = false`

bonkey:
> actually that's why I didn't create a property and created a function to set the completion inside, as there is no need to make it class and then need to handle weak self everywhere 🔫

[Seems like comments disappear after code update.]

Well, still I like a field much more and if you need to do something uncommon like the `func` it feels it's better approach... Don't know which direction is better, yet. Maybe I'll make `[weak self]`s and we can sit over it again?

bonkey:
Seems like `[weak self]` is already around all `.strategyCompletion` calls. Which other places you had in mind?

hani-ibrahim:
I meant each place self is mentioned in the class since it was struct and now class ... not only the completion. (I prefer to always use weak for safety)

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -207,18 +207,18 @@ extension LoggedInSummaryActionHandler {
 
 extension LoggedInSummaryActionHandler {
 
-    fileprivate func updateDataModel(addresses: CheckoutAddresses?, usingCartCheckout: Bool = false) {
+    fileprivate func updateDataModel(addresses: CheckoutAddresses?, byRemovingCartCheckout removeCartCheckout: Bool = true) {
         guard let selectedArticleUnit = dataSource?.dataModel.selectedArticleUnit else { return }
 
         let dataModel: CheckoutSummaryDataModel
-        if usingCartCheckout {
+        if removeCartCheckout {
```
hani-ibrahim: it's now misleading ... `removeCartCheckout` should be by default false and this if condition is switched

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -207,18 +207,18 @@ extension LoggedInSummaryActionHandler {
...
             dataModel = CheckoutSummaryDataModel(selectedArticleUnit: selectedArticleUnit, cartCheckout: cartCheckout, addresses: addresses)
         } else {
             dataModel = CheckoutSummaryDataModel(selectedArticleUnit: selectedArticleUnit, cartCheckout: nil, addresses: addresses)
         }
         delegate?.updated(dataModel: dataModel)
 
-        if usingCartCheckout && cartCheckout?.checkout == nil && hasAddresses {
+        if removeCartCheckout && cartCheckout?.checkout == nil && hasAddresses {
```
hani-ibrahim: ! removeCartCheckout

AtlasUI/AtlasUI/Utils/UIView.swift
hani-ibrahim: it's not about few calls, it's about readability. I don't like these "anchor" thing
.
My code snippet is not final, actually after second thought it could be simply: 
(didn't try it , may need some adjustment) 

``` swift
enum ViewAnchor {
    case top
    case right
    case bottom
    case left

    private func constraint(fromView view1: UIView, toView view2: UIView) -> NSLayoutConstraint {
        switch self {
        case .top: return view1.topAnchor.constraint(equalTo: view2.topAnchor)
        case .right: return view1.rightAnchor.constraint(equalTo: view2.rightAnchor)
        case .bottom: return view1.bottomAnchor.constraint(equalTo: view2.bottomAnchor)
        case .left: return view1.leftAnchor.constraint(equalTo: view2.leftAnchor)
        }
    }

}


func snapAnchorToSuperView(anchor: ViewAnchor, constant: CGFloat = 0) {
    guard let superview = superview else { return }
    translatesAutoresizingMaskIntoConstraints = false
    let constraint = anchor.constraint(fromView: self, toView: superview)
    constraint.constant = constant
    constraint.isActive = true
}
```

bonkey: And the best code is no code. When you create two new methods, both without handling full cases is making the readability much lower.

Second thought is much better. I thought about something like that.

bonkey:
@hani-ibrahim Just naming cleanup, didn't check the correctness fully: ee6976172cd7e1d0a8fbfc78b10102fe9e09bcf5

bonkey:
Accidentally closed when renamed parent branch, so I'll make new PR

PR #576, opened 2016-11-23
#571: Initial Swift 3.0 migration

bonkey:
There're some still some imperfections, but shouldn't influence the whole flow


bonkey:
UI tests are passing locally.

bonkey:
I'm merging to have `swift3-pre` ready.

PR #583, opened 2016-11-29
Code tuning for Swift 3

bonkey:
It's not yet finished, but not much left:

- Verify argument names with “_” in method signatures
- Remove all `Type` from type names
- Verify all `as Any...`
- Rename all extension files for NS*/UI* classes
- Rename all `URL` arguments and properties to lowercase `url`
- Swimat format all changed files (tedious, needs to be manual :-/)
- Verify all @escaping in closures
- Format all guard-let-else in Models
- Review displayErrorFullScreen and rest of displayError
- Review process(forceFullScreenError:)


AtlasUI/AtlasUI/CheckoutSummaryActionHandler.swift
```diff
@@ -26,8 +26,8 @@ protocol CheckoutSummaryActionHandler {
     weak var delegate: CheckoutSummaryActionHandlerDelegate? { get set }
 
     func handleSubmitButton()
-    func showPaymentSelectionScreen()
-    func showShippingAddressSelectionScreen()
-    func showBillingAddressSelectionScreen()
+    func presentPaymentSelectionScreen()
+    func presentShippingAddressSelectionScreen()
```
hani-ibrahim: I think these should be as `handleSubmitButton` because now not all modes display the address screen as guest checkout for example

AtlasUI/AtlasUI/CheckoutSummaryLayout.swift
```diff
@@ -8,7 +8,6 @@ protocol CheckoutSummaryLayout {
 
     var navigationBarTitleLocalizedKey: String { get }
     var showCancelButton: Bool { get }
-    var showPrice: Bool { get }
```
hani-ibrahim: why I'm seeing my updates here ❓

AtlasUI/AtlasUI/CheckoutSummaryViewModel.swift
```diff
@@ -18,22 +18,18 @@ struct CheckoutSummaryViewModel {
 
 extension CheckoutSummaryViewModel {
 
-    fileprivate func validateAgainstOldDataModel(_ oldDataModel: CheckoutSummaryDataModel) {
-        checkPriceChange(oldDataModel)
-        checkPaymentMethod(oldDataModel)
+    fileprivate func validate(against oldDataModel: CheckoutSummaryDataModel) {
+        didPriceChange(comparedTo: oldDataModel)
+        didPaymentClear(comparedTo: oldDataModel)
```
hani-ibrahim: `didPaymentClear` ? this is like a question and the method doesn't return anything so I think the name is miss leading, I would prefer the previous one as the method is handling the price and payment changes.

AtlasUI/AtlasUI/OAuth2LoginViewController.swift
```diff
@@ -45,10 +45,7 @@ final class OAuth2LoginViewController: UIViewController {
         view.backgroundColor = .white
         view.addSubview(webView)
 
-        webView.topAnchor.constraint(equalTo: self.topLayoutGuide.bottomAnchor).isActive = true
-        webView.bottomAnchor.constraint(equalTo: self.bottomLayoutGuide.topAnchor).isActive = true
-        webView.leadingAnchor.constraint(equalTo: self.view.leadingAnchor).isActive = true
-        webView.trailingAnchor.constraint(equalTo: self.view.trailingAnchor).isActive = true
+        webView.bind(toEdges: self)
```
hani-ibrahim: just `fillInSuperview`?

AtlasUI/AtlasUI/Utils/UIView.swift
```diff
@@ -50,14 +50,22 @@ extension UIView {
...
-    func snapAnchorToSuperView(anchor: ViewAnchor, constant: CGFloat = 0) {
+    func bind(toSuperview anchor: ViewAnchor, constant: CGFloat = 0) {
         guard let superview = superview else { return }
         translatesAutoresizingMaskIntoConstraints = false
         let constraint = anchor.constraint(fromView: self, toView: superview)
         constraint.constant = constant
         constraint.isActive = true
     }
 
+    func bind(toEdges viewController: UIViewController) {
```
hani-ibrahim: I don't think it's needed, Just `fillInSuperView` should work

AtlasUI/AtlasUI/CheckoutSummaryActionHandler.swift
bonkey: So, `handleShippingAddressSelection()`, `presentPaymentSelection()` etc?

AtlasUI/AtlasUI/CheckoutSummaryLayout.swift
bonkey: Merge from master

AtlasUI/AtlasUI/CheckoutSummaryActionHandler.swift
hani-ibrahim: yeah or `handleShippingAddressButton` is better I think

AtlasUI/AtlasUI/CheckoutSummaryViewModel.swift
bonkey: True it's not best, but former was meaningless. I spend a couple of cycles to understand what was happening there.

AtlasUI/AtlasUI/Utils/UIView.swift
bonkey: I can check it, but I assumed there was a reason why it was using topLayoutGuide and bottomLayoutGuide instead.

AtlasUI/AtlasUI/CheckoutSummaryViewModel.swift
hani-ibrahim: uhm, then may be the methods return the boolean if it is changed and display the error here instead of the method

AtlasUI/AtlasUI/Utils/UIView.swift
hani-ibrahim: I think that's because it's very old and it was done before creating these helper methods

bonkey: There's a reason:

screen shot 2016-11-29 at 12 38 12

hani-ibrahim:
👍

PR #587, opened 2016-12-05
Fix branch check in “calypso version create”

bonkey:
Just strip new line from branch name. Rest is split to smaller methods.


bonkey:
👍

hani-ibrahim:
👍

Issue #528, opened 2016-11-01
Terms and conditions in registration are always in English


bonkey:
The ToC always refers to zalando.co.uk. It doesn't match selected sales channel.

bonkey:
Fixed with #498

bonkey:
Reported in auth [#62](https://github.bus.zalan.do/Atlas/auth/issues/62), [#63](https://github.bus.zalan.do/Atlas/auth/issues/63)

PR #591, opened 2016-12-07
#528: Added Accept-Language header in AtlasUI

bonkey:
- Default to `AltasUI.interfaceLanguage`.
- Still doesn't show correct language on login webview :-/


bonkey:
👍

AtlasUI/AtlasUI/OAuth2LoginViewController.swift
```diff
@@ -44,7 +44,8 @@ final class OAuth2LoginViewController: UIViewController {
         view.addSubview(webView)
         webView.fillInSuperview()
 
-        webView.loadRequest(URLRequest(url: loginURL))
+        let language = AtlasAPIClient.shared?.config.interfaceLocale.languageCode
```
hani-ibrahim: `interfaceLocale` or country?

bonkey: Interface. It's used for our translations in UI

bonkey: Don't understand the question. It should be. What do you mean?

hani-ibrahim: then the term and condition should be displayed based on the sales channel and not the language of the interface because each country has different terms and conditions

bonkey:
👍

hani-ibrahim:
👍

Issue #595, opened 2016-12-08
Add languages and sales channels in demo app


bonkey:
Add tests to switch and verify ToC

PR #605, opened 2016-12-09
#595: Nice demo app revamp

bonkey:
1. New languages
2. No annoying forced login on profile page
3. Sales channel selection


AtlasDemo/AtlasDemo/InterfaceLanguage.swift
```diff
@@ -0,0 +1,47 @@
...
+    case english = "en"
+    case german = "de"
+    case french = "fr"
+    case italian = "it"
+    case polish = "pl"
+
+    static let all: [InterfaceLanguage] = [.english, .german, .french, .italian, .polish]
+
+    init(index: Int) {
+        self = InterfaceLanguage.all[index]
```
hani-ibrahim: guard for out of range?

AtlasSDK/AtlasSDK/Atlas.swift
```diff
@@ -6,6 +6,14 @@ import Foundation
 
 public typealias AtlasClientCompletion = (AtlasResult<AtlasAPIClient>) -> Void
 
+extension NSNotification.Name {
+
+    public static let AtlasAuthorized = NSNotification.Name(rawValue: "Atlas.NotificationAuthorized")
+    public static let AtlasDeauthorized = NSNotification.Name(rawValue: "Atlas.NotificationDeauthorized")
+    public static let AtlasAuthorizationChanged = NSNotification.Name(rawValue: "Atlas.NotificationAuthorizationChanged")
```
hani-ibrahim: why 3 notifications while you are using only in demo app. I think `AtlasAuthorized / AtlasDeauthorized` OR `AtlasAuthorizationChanged` is enough

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Atlas.swift
bonkey: Well, wanted to make it simple for end developer. Instead of pulling out userInfo from generic (`...Changed`) notification, you can watch for either specific notification, or if you want to handle just a change – like in demo – you have this change only.

AtlasDemo/AtlasDemo/InterfaceLanguage.swift
bonkey: Naah, as then I need to make it optional or lie and return a default language. Too much burden without any real outcome.

PR #609, opened 2016-12-12
#608: Updated translations for 5 languages

bonkey:
#608: Updated translations for 5 languages


bonkey:
👍

bonkey:
👍

AtlasUI/AtlasUI/Utils/Localizer.swift
```diff
@@ -45,7 +51,25 @@ struct Localizer {
...
+        do {
+            return try format(string: key, bundle: self.localizedStringsBundle, formatArguments: formatArguments)
+        } catch let error {
+            if case let Error.missingTranslation(missingKey, language) = error {
+                AtlasLogger.logError("Translation not found for '\(missingKey)' language: \(language)")
+                if Debug.isEnabled {
+                    return key
+                }
+            }
+            return (try? format(string: key, bundle: self.localizedStringsFallbackBundle, formatArguments: formatArguments)) ?? ""
```
hani-ibrahim: instead of empty string ... may be "KEY NOT FOUND"?

AtlasUI/AtlasUI/Utils/Localizer.swift
```diff
@@ -45,7 +51,25 @@ struct Localizer {
...
+        }
+    }
+
+    private func format(string key: String, bundle: Bundle, formatArguments: [CVarArg?]? = nil) throws -> String {
+        let localizedString = NSLocalizedString(key, bundle: bundle, comment: "")
+
+        if localizedString.length == 0 {
```
hani-ibrahim: - One thought: currently most of the missing translations do exists but with the english translation instead of the correct translation. as example: we have `"AtlasAPIError.message.noData" = "Empty response";` in the german localization file. This one is clearly not translated but also this condition `localizedString.length == 0` will not be triggered for this case. So do we have to delete all keys that does not have a correct translation in order for this condition to be triggered?

- Another thing: What should be the process of adding a new key? now we have 5 languages so it would be a long process trying to translate them and this would be time consuming too.

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -88,23 +74,35 @@
...
 "AtlasAPIError.message.checkoutFailed" = "Unknown checkout error";
-
 "AtlasConfigurationError.title" = "Configuration Error";
 "AtlasConfigurationError.message.incorrectConfigServiceResponse" = "Server returned unknown response";
 "AtlasConfigurationError.message.missingClientId" = "Missing client identifier";
 "AtlasConfigurationError.message.missingSalesChannel" = "Missing sales channel";
 "AtlasConfigurationError.message.missingInterfaceLanguage" = "Missing interface language";
+"summaryView.label.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.payWithPayPal" = "Confirm and pay with PayPal";
```
hani-ibrahim: Why moving these 3 down with errors? .... huh Fuzzy?

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -88,23 +74,35 @@
...
+
+/*Guest Summary View*/
+"guestSummaryView.address.add" = "Create new address";
+"guestSummaryView.address.create" = "Create new Address";
+"guestSummaryView.address.update" = "Update Address";
+"addressFormView.emailaddress" = "Email Address";
+"formValidation.vaildEmail" = "Please enter a valid email address";
+"addressListView.add.type.StandardAddressDataModelCreationStrategy" = "Standard";
+"addressListView.add.type.PickupPointAddressDataModelCreationStrategy" = "Pickup Point";
+"addressListView.add.type.AddressBookImportDataModelCreationStrategy" = "Import from Address Book";
```
hani-ibrahim: these 3 `addressListView.add.type` are not related to Guest checkout. They are used in all modes not guest checkout only

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -88,23 +74,35 @@
...
 "AtlasConfigurationError.title" = "Configuration Error";
 "AtlasConfigurationError.message.incorrectConfigServiceResponse" = "Server returned unknown response";
 "AtlasConfigurationError.message.missingClientId" = "Missing client identifier";
 "AtlasConfigurationError.message.missingSalesChannel" = "Missing sales channel";
 "AtlasConfigurationError.message.missingInterfaceLanguage" = "Missing interface language";
+"summaryView.label.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.payWithPayPal" = "Confirm and pay with PayPal";
+
+/*Guest Summary View*/
```
hani-ibrahim: I would prefer to have the translations ordered in a logical order especially that errors should be the last group

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -88,23 +74,35 @@
...
+"guestSummaryView.address.add" = "Create new address";
+"guestSummaryView.address.create" = "Create new Address";
+"guestSummaryView.address.update" = "Update Address";
+"addressFormView.emailaddress" = "Email Address";
+"formValidation.vaildEmail" = "Please enter a valid email address";
```
hani-ibrahim: I know validEmail is **_used_** for guest checkout but it also could be theoretically used in any case since it's related to the forms validation. What if we implemented native register new account form screen instead of the current web view?
I think it needs to be with the other `formValidation`

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -88,23 +74,35 @@
...
 "AtlasConfigurationError.message.missingInterfaceLanguage" = "Missing interface language";
+"summaryView.label.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.guestCheckout" = "Checkout as guest";
+"summaryView.submitButton.payWithPayPal" = "Confirm and pay with PayPal";
+
+/*Guest Summary View*/
+"guestSummaryView.address.add" = "Create new address";
+"guestSummaryView.address.create" = "Create new Address";
+"guestSummaryView.address.update" = "Update Address";
+"addressFormView.emailaddress" = "Email Address";
```
hani-ibrahim: same thing there too, I would like to keep things generic, since they are really generic

AtlasUI/AtlasUI/en.lproj/Localizable.strings
```diff
@@ -1,60 +1,47 @@
...
 "summaryView.label.address.billing" = "Billing Address";
 "summaryView.label.address.packstationAbbreviation" = "P.S.";
 "summaryView.label.emptyAddress.shipping" = "No Shipping Address";
 "summaryView.label.emptyAddress.billing" = "No Billing Address";
 "summaryView.label.estimatedDelivery" = "Estimated Delivery";
 "summaryView.label.unitSize" = "Size: %@";
-"summaryView.label.orderHeader" = "\nWe are happy about your order\nThanks a lot!";
-"summaryView.label.orderNumber" = "Order number";
-"summaryView.label.orderImageSaved" = "✓ Image saved to your photo library";
-"summaryView.button.saveOrderImage" = "Save order details image";
```
hani-ibrahim: these 4 keys related to order screen are missing

AtlasUI/AtlasUI/Utils/Localizer.swift
bonkey: It's case for production, so I was considering `key` actually, but for sure not key not found :)

AtlasUI/AtlasUI/Utils/Localizer.swift
hani-ibrahim: for the first point, I saw that you already cleaned these keys in other languages

bonkey: 1. I have already zeroed them in one of this commits, to make them work with  `== 0`
2. Let's talk, let me go down from the meeting :)

AtlasUI/AtlasUI/en.lproj/Localizable.strings
bonkey: Theoretically POEditor keep them in original... But I'm not sure what's the logic behind it.

screen shot 2016-12-12 at 16 59 49

bonkey:
👍

hani-ibrahim:
👍

bonkey:
👍

hani-ibrahim:
👍

PR #610, opened 2016-12-12
#601: Handle user denied for photos access

hani-ibrahim:
fixing #601


AtlasUI/AtlasUI/CheckoutSummaryOrderStackView.swift
```diff
@@ -99,16 +99,15 @@ extension CheckoutSummaryOrderStackView {
         let (contentOffset, frame) = prepareViewForTakingImage(scrollView: scrollView)
 
         guard let image = scrollView.takeScreenshot() else {
+            UserMessage.displayError(error: AtlasCheckoutError.unclassified)
```
hani-ibrahim: I prefer to return this back because it might fail with the next versions of iOS to get the image Context

hani-ibrahim:
👍

AtlasUI/AtlasUI/CheckoutSummaryOrderStackView.swift
```diff
@@ -99,16 +99,19 @@ extension CheckoutSummaryOrderStackView {
...
             cleanupViewAfterTakingImage(scrollView: scrollView, originalContentOffset: contentOffset, originalFrame: frame)
             return
         }
         cleanupViewAfterTakingImage(scrollView: scrollView, originalContentOffset: contentOffset, originalFrame: frame)
 
-        if !UIApplication.unitTestsAreRunning {
-            UIImageWriteToSavedPhotosAlbum(image, nil, nil, nil)
+        guard !UIApplication.unitTestsAreRunning else {
+            showSavedLabel()
```
bonkey: Oh, please try not to modify existing code one day, just [extend it](https://en.wikipedia.org/wiki/Open/closed_principle)

I mean this `unitTestsAreRunning`... why it should behave differently for tests? I guess simulator, but important is – if it's a mock, don't test it or implement it like there's similar situation in production, not test-based environment.

bonkey:
👍

AtlasUI/AtlasUI/CheckoutSummaryOrderStackView.swift
hani-ibrahim: the problem is not because of simulator. it's because `UIImageWriteToSavedPhotosAlbum` display alert to the user to allow access to photos or not and I can't complete the tests with this alert so I fall back to the success status directly

bonkey:
👍

AtlasUI/AtlasUI/CheckoutSummaryOrderStackView.swift
hani-ibrahim: uhm, okay I'll try

bonkey:
👍

hani-ibrahim:
👍

bonkey:
👍

PR #613, opened 2016-12-13
#612: Let’s try codecov with >=75% and drop max in 1%…

bonkey:
#612: Let’s try codecov with >=75% and drop max in 1%…


hani-ibrahim:
👍

bonkey:
👍
👍

bonkey:
👍

Issue #392, opened 2016-10-04
Change token keychain key to depended on environment parameters


bonkey:
Should be clientId, maybe configuration url, api url and/or redirectUrl from OAuth

Question: if `Atlas.logoutUser()` should clear all authentication tokens or just current one (add function to do it anyway)

bonkey:
Related #493

bonkey:
```
struct APIAccessToken {

    private static let keychainKey = "APIAccessToken-ENV-CLIENTID-MAYBE-LOGINURL"

```

PR #615, opened 2016-12-13
#392: Refactor API token handling

bonkey:
- Moved authorization calls from Atlas struct to AtlasAPIClient
- Extended notification
- Extended keychain key to allow multiple keys for multiple clients and environments
- Unified useSandboxEnvironment and useSandbox naming


bonkey:
Yeah, it's this great pull request, which shows how to make simple things hard... Simple when you keep static and/or singletons. But yes, I know it's lying – in the end it goes to single keychain. So it's only a side-effect. Main reason: multiple tokens for different environments and clients

AtlasSDK/AtlasSDK/RequestBuilder.swift
```diff
@@ -51,8 +55,8 @@ struct RequestBuilder {
     fileprivate func buildRequest() throws -> URLRequest {
         let request: URLRequest = try {
             var r = try URLRequest(endpoint: self.endpoint)
-            if self.endpoint.requiresAuthorization {
-                r.authorize(withToken: APIAccessToken.retrieve())
+            if let token = self.authenticationToken {
```
bonkey: @hani-ibrahim Riddle of the day: this closure is run 3 times on `/customer` call, which is probably not that wrong (but shouldn't), but why on earth only the first call contain value in `self.authenticationToken`? 

```
-- TOKEN --
▿ Optional<String>
  - some : "TOKEN"


🕷  curl -i \
	-X GET \
	-H "Content-Type: application/json" \
	-H "X-Sales-Channel: 82fe2e7f-8c4f-4aa1-9019-b6bde5594456" \
	-H "Authorization: Bearer TOKEN" \
	-H "Accept: application/x.zalando.customer+json,application/x.problem+json" \
	"http://localhost.charlesproxy.com:9080/customer"
-- TOKEN --
nil


-- TOKEN --
nil


🕷  curl -i \
	-X GET \
	-H "Content-Type: application/json" \
	-H "Accept: application/x.zalando.customer+json,application/x.problem+json" \
	-H "X-Sales-Channel: 82fe2e7f-8c4f-4aa1-9019-b6bde5594456" \
	"http://localhost.charlesproxy.com:9080/customer"
```

I'll show you later if I don't catch it earlier.

bonkey: 2 times actually, the second above was some ghost breakpoint

bonkey:
👍

bonkey:
👍

AtlasSDK/AtlasSDK/APIAccessToken.swift
```diff
@@ -6,20 +6,106 @@ import Foundation
...
+
+    var value: String?
+    var key: String {
+        let components = [APIAccessToken.keychainKeyPrefix,
+            clientId,
+            Options.Environment(useSandboxEnvironment: useSandboxEnvironment).rawValue]
+        return components.joined(separator: APIAccessToken.componentsSeparator)
+    }
+
+    fileprivate init(config: Config, value: String? = nil) {
```
hani-ibrahim: Isn't just passing `clientId` and `useSandboxEnvironment` enough? instead of passing the whole `config`

AtlasSDK/AtlasSDK/RequestBuilder.swift
```diff
@@ -5,20 +5,25 @@
...
 struct RequestBuilder {
 
     let endpoint: Endpoint
     let urlSession: URLSession
+    let authenticationTokenProvider: AuthenticationTokenProvider?
 
     fileprivate(set) var taskResponse: DataTaskResponse?
 
-    init(forEndpoint endpoint: Endpoint, urlSession: URLSession = URLSession.shared) {
+    init(forEndpoint endpoint: Endpoint,
```
hani-ibrahim: isn't `init(forEndpoint endpoint: Endpoint, urlSession: URLSession = URLSession.shared, token: String? = nil) {` enough? why making it block?

AtlasSDK/AtlasSDK/APIAccessToken.swift
```diff
@@ -6,20 +6,106 @@ import Foundation
 
 struct APIAccessToken {
 
-    fileprivate static let keychainKey = "APIAccessToken"
+    enum Error: Swift.Error {
```
bonkey: I'd certainly do it for meaningful errors, which go to end-developer. Those are just ad-hoc to throw something more than `Swift.Error`. And based on this reasoning I wouldn't pollute Errors.swift with unmeaningful stuff.

AtlasSDK/AtlasSDK/APIAccessToken.swift
bonkey: It would at this moment, but the fewer arguments, the better. You don't have to change signature if you would add something from config (and I wonder, maybe there should be more, like salesChannel?)

I know, it's coupling to `Config`, but this structure is so tiny and rarely used, that I wouldn't care about it.

AtlasSDK/AtlasSDK/RequestBuilder.swift
bonkey: It was like this for some time today, when I fought with neverending login form appearing after successful login. It's because we create RequestBuilder before authentication token is known, so ew need a dynamic block instead of value.

hani-ibrahim: Oops! but it's not clear that this is the purpose of it.
May be the endpoint could take the config as a parameter instead of sales Channel, client id, etc..
and then we will not have to do all that as we will already have the config and could get the token easily.
I think this would simplify the Endpoints as well.

OR may be make the config as static 👹

AtlasSDK/AtlasSDK/APIAccessToken.swift
hani-ibrahim: then, may be optional init?

bonkey:
👍

hani-ibrahim:
👍

Issue #577, opened 2016-11-28
Localize payment method (Demo feedback)


bonkey:
No translations for payment strings

hani-ibrahim:
available methods: https://github.bus.zalan.do/Atlas/checkout-api/blob/c4d3936f03ab5022d58b7f3f033024299919d17d/src/main/java/de/zalando/atlas/checkout/api/resource/checkout/PaymentMethod.java

PR #631, opened 2016-12-15
#577: Localizing Payment Method

hani-ibrahim:
fixing #577


hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Models/PaymentMethodType.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public enum PaymentMethodType: String {
```
bonkey: Are you sure it's needed, you _Filthy Enumer_? Why to include another code required to modification after every backend change? Can't we just directly translate it as "PAYMENT_METHOD.<KEY>" or sth?

hani-ibrahim: uhm, but it's it good for the user of the SDK to know the available values? using it as string will be opened to the developer imagination of for the possible values 🤔

hani-ibrahim:
👍

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Models/PaymentMethodType.swift
```diff
@@ -0,0 +1,72 @@
+// swiftlint:disable cyclomatic_complexity
```
bonkey: Ha, you wanted to hide it, but It would be killed with `calypso clean` anyway, because it's over copyright :)

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Models/PaymentMethodType.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+//  Copyright © 2016 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public enum PaymentMethodType {
```
bonkey: BTW. That's why I didn't like enum approach with those doubled strings and boilerplate just to print out the translated label

bonkey:
👍

AtlasSDK/AtlasSDK/Models/PaymentMethodType.swift
hani-ibrahim: Oops
It's not the only place where it's used 😧

bonkey:
I can +1 it later :)

hani-ibrahim:
👍

bonkey:
👍

Issue #579, opened 2016-11-28
Implement opening the address list without addresses scenario


bonkey:
Open "Add New Address" and go back to Checkout on success (see [address flow](https://github.com/zalando-incubator/atlas-ios/blob/master/Documentation/Address%20Book%20Flow.png)

PR #632, opened 2016-12-15
#579: No addresses scenario 

hani-ibrahim:
fixing #579


AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -98,34 +95,73 @@ class LoggedInSummaryActionHandler: CheckoutSummaryActionHandler {
...
             let addresses: [EquatableAddress] = userAddresses.map { $0 }
-            let creationStrategy = ShippingAddressViewModelCreationStrategy()
-            let addressViewController = AddressListViewController(initialAddresses: addresses, selectedAddress: self?.shippingAddress)
-            addressViewController.addressUpdatedHandler = { self?.updated(address: $0) }
-            addressViewController.addressDeletedHandler = { self?.deleted(address: $0) }
-            addressViewController.addressSelectedHandler = { self?.select(shippingAddress: $0) }
-            addressViewController.actionHandler = LoggedInAddressListActionHandler(addressViewModelCreationStrategy: creationStrategy)
-            addressViewController.title = Localizer.format(string: "addressListView.title.shipping")
-            AtlasUIViewController.shared?.mainNavigationController.pushViewController(addressViewController, animated: true)
+            if !addresses.isEmpty {
```
bonkey: Maybe better guard + return to present "exceptional situation"? Or if this way, maybe separated funcs for two behaviours to get it read like 

_If there are no adresses, add new address, otherwise..._

bonkey: Oh, actually it is like this... some maybe less generic naming for those funcs?

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -98,34 +95,73 @@ class LoggedInSummaryActionHandler: CheckoutSummaryActionHandler {
...
+                completion(address)
+            }
+            viewController.present()
         }
+        addressCreationStrategy?.execute()
     }
 
 }
 
+// MARK: Handle Confirmation
```
bonkey: I think there should be `swiftlint` for discouraging `// MARK:` 👹

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
```diff
@@ -231,6 +269,7 @@ extension LoggedInSummaryActionHandler {
 
 }
 
+// MARK: Address Modifications
 extension LoggedInSummaryActionHandler {
 
     fileprivate func updated(address: EquatableAddress) {
```
bonkey: How come `swiftlint` is not crying yet at this line count?

hani-ibrahim: power of extensions 😎

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
bonkey: Clear and easy as `UITableViewCell` ;-)

https://twitter.com/arekholko/status/808668358405132289

bonkey: BTW. Maybe `// MARK:-` with **–**

AtlasUI/AtlasUI/LoggedInSummaryActionHandler.swift
hani-ibrahim: Guard is good but I don't see any benefits here, and sometimes it makes reading difficult. There are no algorithms or logic that will be created if guard passed, it just different behavior based on different cases. it's not really something like if this happen then fallback to the exceptional case. It's not that exceptional to use guard it's a very possible scenario.

Also the code will be more unreadable for me if written in this way:
``` swift
guard !addresses.isEmpty else {
    self?.showFormViewController(strategy: ShippingAddressViewModelCreationStrategy()) { [weak self] newAddress in
        self?.select(shippingAddress: newAddress)
    }
    return
}

self?.showAddressListViewController(forShippingAddressWithAddresses: addresses)
```

bonkey:
👍

hani-ibrahim:
👍

bonkey:
👍

hani-ibrahim:
👍

PR #635, opened 2016-12-16
#624: Use currency from backend in price formatter

bonkey:
- Refactored code to use Money instead of MoneyAmount and Price
- Refactored shipping price to be dynamic based on totalPrice


bonkey:
👍

bonkey:
👍

AtlasSDK/AtlasSDK/Models/Price.swift
```diff
@@ -1,37 +0,0 @@
-//
-//  Copyright © 2016 Zalando SE. All rights reserved.
-//
-
-import Foundation
-
-public struct Price {
```
hani-ibrahim: Nice 👍

AtlasUI/AtlasUI/Utils/Localizer.swift
```diff
@@ -106,8 +107,12 @@ extension Localizer {
         return shared.format(string: string, formatArguments: formatArguments)
     }
 
-    static func format(price: NSNumber) -> String? {
-        return shared.format(price: price)
+    static func format(price: NSNumber, withCurrency currencySymbol: String? = nil) -> String? {
```
hani-ibrahim: why we have two method with the same signature but one is static and one is not?

hani-ibrahim:
👍

AtlasUI/AtlasUI/Utils/Localizer.swift
bonkey: Just to not write `shared` in `Localizer.shared.format(...` everytime

hani-ibrahim: then may be remove the non static one and add the logic directly in the static one?

bonkey:
👍

AtlasUI/AtlasUI/Utils/Localizer.swift
bonkey: I don't want to castrate real class with it, these statics are just syntax sugar

hani-ibrahim:
👍

PR #653, opened 2017-01-02
#652: Added keyboardType in AddressFormField

bonkey:
#652: Added keyboardType in AddressFormField


bonkey:
@hani-ibrahim this stupid `testPlaceOrder`... might be probably faster if we sit over together over it. It works on my machine, but fails often on CIs, so I created #655 and we can merge this PR in spite of travis' problems with it

bonkey:
👍

hani-ibrahim:
👍

Issue #557, opened 2016-11-15
"Add address" crashes on iPad


bonkey:
> Terminating app due to uncaught exception 'NSGenericException', reason: 'Your application has presented a UIAlertController (<UIAlertController: 0x7fdf2b658870>) of style UIAlertControllerStyleActionSheet. The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.'

<img width="880" alt="screen_shot_2016-11-15_at_11_39_15" src="https://cloud.githubusercontent.com/assets/39100/20302628/4cc75b66-ab28-11e6-887c-81a62eab489d.png">

bonkey:
@hani-ibrahim I think it could be still open.

PR #681, opened 2017-01-13
#557: Showing alerts instead of action sheets on iPad

bonkey:
#557: Showing alerts instead of action sheets on iPad


bonkey:
👍

AtlasSDK/AtlasSDK/Extensions/UIDevice.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  Copyright © 2017 Zalando SE. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+extension UIDevice {
```
bonkey: Yeah, could be there as well, as it goes to Extensions. But it's anyway `UI`

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Extensions/UIDevice.swift
bonkey: Ah, you ask why in AtlasSDK? Well, because it's `Extensions` – default folder is in SDK

hani-ibrahim: but we have `Utils` folder in AtlasUI for that

bonkey: I created shared Extensions... Well, maybe Utils is better location actually, as _it's anyway `UI`_

bonkey:
👍

bonkey:
This could be problem with zappr statuses:

> zappr — connect ETIMEDOUT 192.30.253.117:443

hani-ibrahim:
👍

PR #685, opened 2017-01-17
#682: Let's have own JSON parser

bonkey:
1. Yes, tests are not compiling, yet. For the sake of simplicity I assumed it will be only deserializing  read-only parser, but found out that in tests for Guest checkout it's modified: https://github.com/zalando-incubator/atlas-ios/blob/issue-682/AtlasSDK/AtlasSDKTests/ConfigTests.swift#L52
1. I'm not very convinced about this approach, but still like it much more than this crazy SwiftyJSON
1. I kept interface as closest as possible to SwiftyJSON, however I don't like everything about it
1. Freddy is just quick solution to make it working. If you like new parser I would consider adding it the same way as `Extensions` to both projects
1. Yeah, no tests for the parser – YET
1. Feel free to criticize a lot – it's not very polished, I wanted to get your impression


AtlasSDK/AtlasSDK/JSON/JSONSubscript.swift
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    subscript(path: JSONSubscript...) -> JSON {
+        get {
+            return self[path] ?? JSON.null
+        }
+    }
+
+    subscript(path: [JSONSubscript]) -> JSON? {
```
hani-ibrahim: is it used?

AtlasSDK/AtlasSDK/ResponseParser.swift
```diff
@@ -18,13 +18,13 @@ struct ResponseParser {
             return completion(.failure(AtlasAPIError.noData))
         }
 
-        let json = JSON(data: data)
+        let json = (try? JSON(data: data)) ?? nil // swiftlint:disable:this redundant_nil_coalescing
```
hani-ibrahim: why `?? nil` ?

AtlasSDK/AtlasSDK/JSON/JSON.swift
```diff
@@ -0,0 +1,107 @@
...
+//
+
+// Heavily influenced by [SwiftyJSON](https://github.com/SwiftyJSON/SwiftyJSON)
+
+import Foundation
+
+struct JSON: CustomStringConvertible {
+
+    enum DataType {
+        case number, string, bool, array, dictionary, null
```
bonkey: They're not standard data types of JSON and they're used only in `internalObject` to select proper `raw...` properties.

AtlasSDK/AtlasSDK/JSON/JSON.swift
```diff
@@ -0,0 +1,105 @@
...
+    private static let decimalNumberType = String(describing: type(of: NSNumber(value: 1)))
+
+    private(set) var type: DataType = .null
+    private(set) var rawArray: [Any] = []
+    private(set) var rawDictionary: [String: Any] = [:]
+    private(set) var rawString: String = ""
+    private(set) var rawNumber: NSNumber = 0
+    private(set) var rawBool: Bool = false
+
+    fileprivate var internalObject: Any {
```
bonkey: Yeah, I'll try to play around it more or less in this way, but as a next step, if you don't mind.

hani-ibrahim: okay 👍

AtlasSDK/AtlasSDK/JSON/JSON.swift
```diff
@@ -0,0 +1,107 @@
...
+            switch newValue {
+            case let number as NSNumber:
+                let newValueType = String(describing: type(of: newValue))
+                if newValueType == JSON.decimalNumberType {
+                    self.type = .number
+                    self.rawNumber = number
+                } else if let boolValue = newValue as? Bool {
+                    self.type = .bool
+                    self.rawBool = boolValue
+                }
```
bonkey: Good catch, however – to be simplified anyway with the former comment.

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
```diff
@@ -0,0 +1,141 @@
...
+    var array: [JSON]? {
+        guard self.type == .array else { return nil }
+        return self.rawArray.flatMap { JSON($0) } // TODO: shouldn't be flatMap
+    }
+
+    var arrayValue: [JSON] {
+        return array ?? []
+    }
+
+    var arrayObject: [Any] {
```
bonkey: Left over from SwiftyJSON. `dictionaryObject` is used, so I kept it consistent. Probably to be simplified anyway.

AtlasDemo/AtlasDemo/DemoArticle.swift
```diff
@@ -103,51 +107,53 @@ extension DemoArticle {
...
     }
 
 }
 
 extension DemoArticle.Attribute {
 
     init?(json: JSON) {
-        guard let name = json["name"].string else { return nil }
+        guard let name = try? json.getString(at: "name"),
+            let values = try? json.getArray(at: "values")
```
bonkey: True. I thought they were mandatory implicitly, as I expected `arrayValue` would crash if they weren't available, but it's not true.

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
```diff
@@ -0,0 +1,141 @@
...
+            d[key] = JSON(value)
+        }
+        return d
+    }
+
+    var dictionaryValue: [String: JSON] {
+        return self.dictionary ?? [:]
+    }
+
+    var dictionaryObject: [String: Any]? {
```
bonkey: `.metadata` in 

* `AtlasSDK/AtlasSDK/Models/GuestPaymentMethod.swift`
* `AtlasSDK/AtlasSDK/Models/PaymentMethod.swift`

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
```diff
@@ -0,0 +1,141 @@
...
+protocol JSONObjectifier {
+
+    init?(json: JSON?)
+
+}
+
+extension URL: JSONObjectifier {
+
+    init?(json: JSON?) {
+        guard let string = json?.string else { return nil }
```
bonkey: `json?.string` checks for it.

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
```diff
@@ -0,0 +1,141 @@
...
+        guard let number = json?.rawNumber, json?.type == .number else { return nil }
+        self = number.floatValue
+    }
+
+}
+
+extension Date: JSONObjectifier {
+
+    init?(json: JSON?) {
+        let dateFormatter = RFC3339DateFormatter()
```
bonkey: We were creating it anyway in models. Extension can't have it, so I didn't want to polute main `JSON` type with it.

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
```diff
@@ -0,0 +1,141 @@
...
+
+        guard let string = json?.rawString,
```
bonkey: But when we use `.string` it's not needed any more ;-)

AtlasSDK/AtlasSDK/JSON/JSONSubscript.swift
bonkey: In `subscript(path: JSONSubscript...)`

bonkey: It allows using `json["level1", "level2", "level3", 0, 0]`

AtlasSDK/AtlasSDK/ResponseParser.swift
bonkey: Because it's `try? Optional<JSON>`, which expands to `??`. But I'll try one thing... while I added `JSON.null`

AtlasSDK/AtlasSDK/JSON/JSONObjectifier.swift
hani-ibrahim: hmm, interesting, actually `metadata` is not used in any place  
also I found this in the Mock json files
``` swift
"metadata": {
   "DEPRECATED_cpm_id": "8255569"
}
```
probably it could be removed but anyway it's out of scope now, I created a note for it in the Technical Debt board

AtlasDemo/AtlasDemo/DemoArticle.swift
```diff
@@ -103,51 +107,52 @@ extension DemoArticle {
...
 }
 
 extension DemoArticle.Attribute {
 
     init?(json: JSON) {
-        guard let name = json["name"].string else { return nil }
+        guard let name = try? json.getString(at: "name") else { return nil }
         self.name = name
-        self.values = json["values"].arrayValue.flatMap { $0.string }
+        let values = (try? json.getArray(at: "values")) ?? []
```
hani-ibrahim: isn't this similar to `let json = (try? JSON(data: data)) ?? nil // swiftlint:disable:this redundant_nil_coalescing`?

hani-ibrahim:
👍

hani-ibrahim:
👍

bonkey:
👍

bonkey:
So, my pleasure to merge it. Hopefully won't introduce _big_ problems 🤡

Issue #686, opened 2017-01-17
Simplify internal JSON converter


bonkey:
Drop `raw...` and `type` and do direct casting for properties instead.

PR #687, opened 2017-01-17
#686: Even less swifty JSON helper

bonkey:
How about now?

#686: Unified and simplified array and dictionary in JSON
#686: Simplified JSON: removed raw… values and JSONObjectifier
#686: Optimized JSON.isBool


bonkey:
👍

bonkey:
@hani-ibrahim 

Unfortunately, `testBillingAddressWithNoAddresses` and `testShippingAddressWithNoAddresses` are failing on my side as well.

And unfortunately, the tests are complex enough to not be easily debugged with just a quick look, so it would be great if you spot the reason.

bonkey:
And Buddybuild passed because... it seems like it didn't run tests. There's

```
    Oops! There were test failures.
```

But no test logs. Argh!

bonkey:
Maybe it's somehow related to the latest addition for [testing frameworks](https://www.buddybuild.com/blog/ios-framework-library-support)?

AtlasSDK/AtlasSDKTests/JSONNumbersTests.swift
```diff
@@ -0,0 +1,17 @@
...
+
+import XCTest
+import Nimble
+
+@testable import AtlasSDK
+
+class JSONNumbersTests: JSONTestCase {
+
+    func testInt0() {
+        expect(0).to(equalJson(path: "numbers", "int0"))
```
bonkey: @hani-ibrahim Compare it to your tests 😎

hani-ibrahim: 🙄

hani-ibrahim:
@bonkey `testBillingAddressWithNoAddresses` is failing because as I said in my first comment, returning nil if the array is empty will be handled as invalid response so these **simple** two tests were failing because they were checking testBillingAddressWith**_NoAddresses_**

hani-ibrahim:
But it's now much better ;)

AtlasSDK/AtlasSDK/AtlasAPIClient.swift
```diff
@@ -39,8 +39,10 @@ public struct AtlasAPIClient {
 
     func fetch<Model: JSONInitializable>(from endpoint: Endpoint, completion: @escaping APIResultCompletion<[Model]>) {
         call(endpoint: endpoint, completion: completion) { response in
-            guard let json = response.body, let jsons = json.array.flatMap({ $0 }) else { return nil }
-            return jsons.flatMap { Model(json: $0) }
+            guard let json = response.body,
+                !json.array.isEmpty
```
bonkey: I expected something like this. Assumed in hurry, that optional array + `flatMap()` == non-optional array + `!isEmpty`.

Fixed. And tests are passing.

hani-ibrahim:
👍

bonkey:
:+1 :

bonkey:
Finally we can see all green...

Issue #671, opened 2017-01-10
Handle normalized address response


bonkey:
Should present both:

* address typed in by the user
* normalized address from the backend

and allow a user to choose

PR #692, opened 2017-01-20
#671: Normalized Address

hani-ibrahim:
fixing #671


AtlasUI/AtlasUI/AddressCheckViewController.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.title = Localizer.format(string: "addressCheckView.title")
+    }
+
+    private dynamic func cancelButtonPressed() {
+        finish(withResult: .cancel)
+    }
+
+    private dynamic func submitButtonPressed() {
+        guard let selectedAddress = rootStackView.selectedAddress else {
+            finish(withResult: .cancel)
```
bonkey: Maybe single-liner?

```
guard let selectedAddress = rootStackView.selectedAddress else { return finish(withResult: .cancel) }
```

AtlasUI/AtlasUI/AddressFormDataModel.swift
```diff
@@ -133,3 +133,20 @@ extension CheckAddress {
...
 
 }
+
+extension AddressFormDataModel {
+
+    fileprivate var isNormalAddress: Bool {
+        return pickupPointId == nil && pickupPointMemberId == nil
+    }
+
+    func update(forCheckAddress address: CheckAddress) {
```
bonkey: I prefer short words in signatures, especially when the type is already in the argument, and additionally: argument itself is from a variable named after the type.

So, how about:

```
func update(from address: CheckAddress) { ...
```

(However, I'd keep the meaning in the keyword, if it's a generic type as `String` and `Int`).

AtlasUI/AtlasUI/AddressFormViewController.swift
```diff
@@ -133,4 +133,8 @@ extension AddressFormViewController: AddressFormActionHandlerDelegate {
         }
     }
 
+    func updateView(withDataModel dataModel: AddressFormDataModel) {
```
bonkey: Ditto (notice dropping "`View`" from function name, as I expect 

```
   func updateView(with dataModel: AddressFormDataModel) { ...
```

AtlasUI/AtlasUI/LoggedInCreateAddressActionHandler.swift
```diff
@@ -10,8 +10,20 @@ struct LoggedInCreateAddressActionHandler: AddressFormActionHandler {
...
     func submit(dataModel: AddressFormDataModel) {
-        validateAddress(dataModel: dataModel) { success in
-            guard let request = CreateAddressRequest(dataModel: dataModel), success else {
+        validateAddress(dataModel: dataModel) { result in
+            guard case .selectAddress(let selectedAddress) = result else {
+                if case .editAddress(let editAddress) = result {
+                    dataModel.update(forCheckAddress: editAddress)
+                    self.delegate?.updateView(withDataModel: dataModel)
+                }
+                self.delegate?.addressProcessingFinished()
```
bonkey: `return self.delegate?.addressProcessingFinished()`?

AtlasUI/AtlasUI/LoggedInCreateAddressActionHandler.swift
```diff
@@ -10,8 +10,20 @@ struct LoggedInCreateAddressActionHandler: AddressFormActionHandler {
...
+                return
+            }
+
+            dataModel.update(forCheckAddress: selectedAddress)
+            self.delegate?.updateView(withDataModel: dataModel)
+
+            guard let request = CreateAddressRequest(dataModel: dataModel) else {
                 self.delegate?.addressProcessingFinished()
```
bonkey: `return self.delegate?.addressProcessingFinished()`?

AtlasSDK/AtlasSDK/Models/CheckAddress.swift
```diff
@@ -52,3 +52,13 @@ extension CheckAddress: JSONRepresentable {
     }
 
 }
+
+public func === (lhs: CheckAddress?, rhs: CheckAddress?) -> Bool {
+    guard let lhs = lhs, let rhs = rhs else { return false }
```
hani-ibrahim: It's used in one case only which is here: https://github.com/zalando-incubator/atlas-ios/blob/issue-671/AtlasUI/AtlasUI/AddressCheckStackView.swift#L34

and actually I need it to be false if both are nil, not sure if this case would ever happen though

bonkey: So, maybe one step earlier? Otherwise I would take it as a small lie for two nils...

bonkey:
👍

AtlasUI/AtlasUI/AddressCheckViewController.swift
hani-ibrahim: Aha! ... now you like one long line :P 
you should remove the 140 char swiftlint 👹

But this is still less then 140 😞

bonkey: Unless it's not exceeding 140 chars 🤡

AtlasUI/AtlasUI/LoggedInCreateAddressActionHandler.swift
hani-ibrahim: not possible because delegate is optional which means it could return nil and the function should not return anything

hani-ibrahim:
👍

AtlasUI/AtlasUI/AddressCheckStackView.swift
```diff
@@ -29,9 +29,14 @@ class AddressCheckStackView: UIStackView {
     }()
 
     var addressesRow: [(label: UILabel, view: AddressCheckRowView)] = []
-    var selectedAddress: CheckAddress? {
+    var selectedAddress: AddressCheck? {
         didSet {
-            addressesRow.forEach { $0.view.selectRow(selected: $0.view.selectButton.address === selectedAddress) }
+            addressesRow.forEach {
+                guard let viewAddress = $0.view.selectButton.address, let selectedAddress = selectedAddress else {
+                    return $0.view.selectRow(selected: false)
```
bonkey: Codecov plugin tells me there're no tests for this block. Probably not that important, I'm just even more sensitive after the journey through JSON tests.

AtlasUI/AtlasUITests/AddressCheckViewControllerTests.swift
```diff
@@ -0,0 +1,135 @@
...
+        self.street = street
+        self.additional = nil
+        self.zip = "10178"
+        self.city = "Berlin"
+        self.countryCode = "DE"
+    }
+
+}
+
+private func == (lhs: AddressCheckResult, rhs: AddressCheckResult) -> Bool {
```
bonkey: Codecov tells no tests, which may mean it's not used.

bonkey:
👍

AtlasUI/AtlasUI/AddressCheckStackView.swift
hani-ibrahim: it's just an else of guard, but actually this condition should never happen, it's because the properties are optional and I need to unwrap them

AtlasUI/AtlasUITests/AddressCheckViewControllerTests.swift
hani-ibrahim: good catch, I used them at first and then not
removed them 👍

hani-ibrahim:
👍

bonkey:
👍

PR #698, opened 2017-01-25
Update .swiftlint.yml

hani-ibrahim:
These warning are only in Demo app because we don't have IBOutlets in the SDK. Also they are not working properly because they are already `fileprivate` but not `private`


hani-ibrahim:
👍

Issue #700, opened 2017-01-26
Move out all binaries from main repo to Git LFS


bonkey:
https://git-lfs.github.com/

bonkey:
Nah, seems problematic. After couple of tries ended up with

```
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
fatal error: unexpected signal during runtime execution
[signal 0xb code=0x1 addr=0x57bdd0f8 pc=0xb8be5]
```

bonkey:
The crashes above were from some ancient version of `git-lfs` binary sitting in `/usr/local/bin`. After forced linking from brew everything went back to normal.

bonkey:
Closing it right away. Finally everything worked, but the pricing model for LFS, even on open source is a blocker here.

PR #707, opened 2017-01-31
#700: Quick try to switch to git-lfs

bonkey:
Another try, as I saw it has support in Tower, so it was quick clickety-click to see if it works.

@hani-eloksh - could you check out this version and see if it has correct binaries on your local machine?


hani-ibrahim:
tried to run the app but got this error

screen shot 2017-02-02 at 10 32 11 am

Then I tried `brew install git-lfs` but it gave me the same errors
Then I tried `git lfs install` which I took from the travis file but I got 
```
The filter.lfs.smudge attribute should be "git-lfs smudge -- %f" but is "git-lfs smudge %f"
Run `git lfs install --force` to reset git config.
``` 

Then I just run `./calypso.rb deps build` and it worked fine 👍

hani-ibrahim:
👍

bonkey:
But using `calypso deps build` is not the way to go, as you override those files, which sit as references, and should be replaced by git-lfs to binaries.

Let sit over it on Friday.

bonkey:
👍

bonkey:
Closing it right away. Finally everything worked, but the pricing model for LFS, even on open source is a blocker here.

PR #709, opened 2017-02-01
#708: Maintenance update of libraries and code

bonkey:
#708: Maintenance update of libraries and code


AtlasSDK/AtlasSDK/Utils/Keychain.swift
```diff
@@ -13,14 +13,15 @@ struct Keychain {
 
     @discardableResult
     static func delete(key: String) -> Bool {
+        let status: OSStatus
```
hani-ibrahim: Actually I was surprised that `defer` sees the variable created after it, now it makes more sense

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Utils/Keychain.swift
bonkey: Me either, but when it worked I kept it.

I'm just thinking (only philosophically, without checking reasoning or internal mechanisms) if it did work, why change? Don't they go [too much into restrictions](http://blog.cleancoder.com/uncle-bob/2017/01/11/TheDarkPath.html)? 👹

bonkey:
👍

bonkey:
Merging without codecov – CI didn't fail, but don't need to play with it any longer.

Issue #593, opened 2016-12-07
Unify internal naming of Pickup points and Packstations


bonkey:
They're two different things. They need to be named, translated, and handled separately and probably differently in different countries.

bonkey:
Needs changes in backend to tell what kind of delivery is available.

bonkey:
Handling different countries in separated topic: #673

bonkey:
For Germany we need to rename everything to Packstation. Pickup point is not supported yet.

PR #735, opened 2017-02-10
#593: Renaming Packstation

hani-ibrahim:
fixing #593


AtlasSDK/AtlasSDK/Models/PickupPoint.swift
```diff
@@ -6,7 +6,18 @@ public struct PickupPoint {
 
     public let id: String
     public let name: String
-    public let memberId: String
+    public let memberId: String?
+
+}
+
+public enum PickupPointType {
```
bonkey: How about instead of long type name, which is not used alone, embed it into parent `PickupPoint`:

```
struct PickupPoint {
// ...
  struct Type {
  // ...
  }
}

hani-ibrahim:
👍

bonkey:
👍

Issue #730, opened 2017-02-09
Fix Cancel from login for the new design


hani-ibrahim:
Cancel from login -> back to size is not responding

PR #741, opened 2017-02-10
#730: Fixing cancel the login screen for new design

hani-ibrahim:
fiing #730


hani-ibrahim:
@bonkey not sure what to test here
it's just a matter of when to hide the collection view.

- previously I was hiding the collection view directly when the user selecting an item 
- now for the initial (unselected) state only, I'm hiding the view after successful transition to a new layout, so if the anything fails it will return back to the collection view opened

bonkey:
👍

hani-ibrahim:
👍

Issue #715, opened 2017-02-06
Present recommendations on order confirmation screen


bonkey:
Will solve #654

bonkey:
Related to #750

PR #745, opened 2017-02-16
#715: Recommendation

hani-ibrahim:
fixing #715 

- [x] Let's sit over `CheckoutResult` cases
- [x] Don't forget about translations
- [x] Add `completion` description in `AtlasUI.presentCheckout()`
- [x] Constraint warnings


AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -88,6 +109,15 @@ final public class AtlasUI {
         viewController.present(atlasUIViewController, animated: true, completion: nil)
     }
 
+    func deregisterResult() {
+        injector.deregister(AtlasUI.Result.self)
```
hani-ibrahim: @bonkey I didn't like this:
I need to save the state of the checkout process (orderPlaced, fullScreen error displayed, etc ) to be able to return that state to the host app
- so I used `injector` for this but this found that the injector is not cleared after the checkout view is dismissed
- may be `AtlasUIViewController` is better, but I don't feel it's a good place to save it, or it is?

AtlasDemo/AtlasDemo/ProductCollectionViewCell.swift
```diff
@@ -31,9 +31,28 @@ class ProductCollectionViewCell: UICollectionViewCell {
     @IBAction func buyNowButtonTapped(_ sender: AnyObject) {
         guard let rootController = UIApplication.shared.keyWindow?.rootViewController, let article = self.article else { return }
         do {
-            try AppSetup.atlas?.presentCheckout(onViewController: rootController, forSKU: article.id)
+            try AppSetup.atlas?.presentCheckout(onViewController: rootController, forSKU: article.id) { [weak self] result in
+                switch result {
+                case .orderPlaced: print("AtlasUI Finished with: Order Placed")
+                case .orderPlacedAndRecommendedItemChosen(let sku): self?.displayRecommendedProduct(sku: sku)
+                case .userCancelled: print("AtlasUI Finished with: User Cancelled")
+                case .errorDisplayed(let error): print("AtlasUI Finished with: Error Displayed(\(error))")
```
bonkey: prints, not UserMessages?

AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
```diff
@@ -138,6 +141,11 @@ extension AtlasAPIClient {
         fetch(from: endpoint, completion: fetchCompletion)
     }
 
+    public func articleRecommendation(withSKU sku: String, completion: @escaping RecommendationCompletion) {
```
bonkey:
```swift
func recommendations(forSKU sku: String, completion: @escaping RecommendationCompletion) { ...
```

AtlasSDK/AtlasSDK/Models/Media.swift
```diff
@@ -0,0 +1,51 @@
...
+//  Copyright © 2017 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public struct Media {
+
+    public let images: [Image]
+
+    public struct Image {
```
bonkey: Maybe in extension?

```
extension Media {

    struct Image { }

}
```

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -7,12 +7,13 @@ import UIKit
 import AtlasSDK
 
 public typealias AtlasUICompletion = (AtlasResult<AtlasUI>) -> Void
```
bonkey: BTW. Let's rename it to `AtlasUIConfigureCompletion`

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -22,9 +23,24 @@ final public class AtlasUI {
...
     }
 
+    /// Result that can returned after presenting the Checkout screen
+    ///
+    /// - orderPlaced: The customer successfully placed the order
+    /// - orderPlacedAndRecommendedItemChosen: The customer successfully placed the order and is interested to view this item 
+    ///                                        (Please open the product detail page for the given SKU)
+    /// - userCancelled: The user cancelled the checkout process
+    /// - errorDisplayed: Error displayed to the user
+    public enum Result {
```
bonkey: Rather `CheckoutResult`

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -22,9 +23,24 @@ final public class AtlasUI {
...
+        case orderPlaced
+        case orderPlacedAndRecommendedItemChosen(sku: String)
```
bonkey: I'd rather have one case for order placed to not handle both cases as an end-developer.

Not sure, but maybe `case orderPlaced(recommendedSKU: String?)`. Would be so much simpler to have 2 different signals, where you could subscribe for `orderPlaced` event and `recommendationSelected` instead of enum with cases.

bonkey: OK, closure would give similar behaviour, but then we should have whole interface for end-dev built over such behaviours.

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -22,9 +23,24 @@ final public class AtlasUI {
...
+        case userCancelled
+        case errorDisplayed(error: AtlasError)
```
bonkey: Not clear if it was only displayed or just finished with such error

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -71,12 +87,17 @@ final public class AtlasUI {
     ///   - viewController: The controller in which AtlasUI will be presented over it
     ///   - sku: The SKU for the item that the user want to buy
     /// - Throws: notInitialized is thrown if this method is called before initializing AtlasUI be calling configure(options:completion:)
-    public func presentCheckout(onViewController viewController: UIViewController, forSKU sku: String) throws {
+    public func presentCheckout(onViewController viewController: UIViewController,
+                                forSKU sku: String,
+                                completion: @escaping AtlasUICheckoutResultCompletion) throws {
```
bonkey: `...AtlasUICheckoutResultCompletion? = nil`

AtlasUI/AtlasUI/AtlasUI.swift
bonkey: Well, probably it's just in a few places, so maybe just return it and propagate, not really store?

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -88,6 +109,15 @@ final public class AtlasUI {
         viewController.present(atlasUIViewController, animated: true, completion: nil)
     }
 
+    func deregisterResult() {
+        injector.deregister(AtlasUI.Result.self)
+    }
+
+    func dismissAtlasCheckoutUI() throws {
+        completion?(try provide())
```
bonkey: better name the result of `provide()` - I thought for a longer moment, what it could be.

AtlasUI/AtlasUI/CheckoutPresentationController.swift
```diff
@@ -81,6 +81,11 @@ final class CheckoutPresentationController: UIPresentationController {
 
     @objc
     fileprivate func dimmingViewTapped() {
-        presentingViewController.dismiss(animated: true, completion: nil)
+        let registeredResult = try? AtlasUI.shared().provide() as AtlasUI.Result
+        if registeredResult == nil {
+            try? AtlasUI.shared().register { AtlasUI.Result.userCancelled }
+        }
+        try? AtlasUI.shared().dismissAtlasCheckoutUI()
```
bonkey: Oh, you hard try here

bonkey:
>  **codebeat**: 2 issues resolved and 12 introduced

bonkey:
👍

AtlasUI/AtlasUI/AtlasUI.swift
hani-ibrahim: I would love to propagate it. But the main (probably the only) problem is that when dismissing the dimming view (the one that is covering the demo app). 
This dimming view is completely outside the checkout view and doesn't know the current state of the checkout process whether the order is placed or an error is displayed. 😤

AtlasDemo/AtlasDemo/ProductCollectionViewCell.swift
hani-ibrahim: It's demo app, thought that alert while be annoying while testing. Imagine you open the checkout and then cancel, an error will be displayed :(

AtlasUI/AtlasUI/AtlasUI.swift
hani-ibrahim: It's not configure, it's the completion of the checkout process itself.

AtlasUI/AtlasUI/AtlasUI.swift
hani-ibrahim: I'm not quite sure that I got your point. Let's set over it when you return back

AtlasUI/AtlasUI/AtlasUI.swift
hani-ibrahim: The whole enum is called only when the view is dismissed, so it's for `finished` state
Probably `finishedWithError` is better

AtlasUI/AtlasUI/AtlasUI.swift
hani-ibrahim: But I think the end developer should handle the completion properly and we should not give the option for not handling anything, especially if the user chose a recommended item. This should (probably must) be handled by the host app.

AtlasUI/AtlasUI/CheckoutPresentationController.swift
hani-ibrahim: 😂

AtlasSDK/AtlasSDK/Models/Media.swift
hani-ibrahim: This is not the case for other model (for example `Article`)  🤔

AtlasUI/AtlasUI/AtlasUI.swift
bonkey: Incorrect. The next line is about the checkout. This [one is in configure](https://github.com/zalando-incubator/atlas-ios/pull/745/files#diff-0a93f8386bf40e7291a7b25d4535a3cfL47), and while we have 2 different aliases, let's differentiate them

hani-ibrahim: Ah I thought you are commenting on the new completion 👍

AtlasSDK/AtlasSDK/Models/Media.swift
bonkey: Well, maybe let's do this with other, too? Not strong reason about it, just I was lost having just one small field for images and bigger `Image` type which eclipses it in the code block.

AtlasDemo/AtlasDemo/ProductCollectionViewCell.swift
bonkey: I thought about error case the most, but if it's not the case – just kill all maybe, but `UserMessage` in handling error below might be useful.

AtlasUI/AtlasUI/AtlasUI.swift
bonkey: - [ ] let's sit over

AtlasUI/AtlasUI/AtlasUI.swift
bonkey: Oh, if it **must** be, why it's not described in the comment above?

Got your point, but still – I might don't want to do it as end-developer (could be for some reason too complex to open whole different controller tree to present another article). In such case recommendations should be optional.

hani-ibrahim:
👍

hani-ibrahim:
👍

bonkey:
👍

hani-ibrahim:
will need to wait a little bit for this PR.
The changes in the catalog API is missed for the recommendation endpoint. So I need to wait for these changes to be applied for recommendation endpoint also 🐼

hani-ibrahim:
👍

hani-ibrahim:
@bonkey I'll not approve any PR before merging this one 😈

hani-ibrahim:
👍

bonkey:
👍

bonkey:
👍

Issue #727, opened 2017-02-08
UICollectionViewFlowLayout is not defined


bonkey:
On [buddybuild](https://dashboard.buddybuild.com/apps/57a305cb34a9450100595b71/build/589b0233703723010092d43a#logs), but I think I saw it too in Xcode earlier.

```
[AtlasUITestsHostsApp] The behavior of the UICollectionViewFlowLayout is not defined because:
[AtlasUITestsHostsApp] the item height must be less than the height of the UICollectionView minus the section insets top and bottom values, minus the content insets top and bottom values.
[AtlasUITestsHostsApp] The relevant UICollectionViewFlowLayout instance is <UICollectionViewFlowLayout: 0x7ffe61cd9520>, and it is attached to <AtlasUI.CheckoutSummaryArticleSelectCollectionView: 0x7ffe620ac000; baseClass = UICollectionView; frame = (0 0; 320 50); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x6080004515e0>; layer = <CALayer: 0x6080004315e0>; contentOffset: {0, 0}; contentSize: {0, 0}> collection view layout: <UICollectionViewFlowLayout: 0x7ffe61cd9520>.
[AtlasUITestsHostsApp] Make a symbolic breakpoint at UICollectionViewFlowLayoutBreakForInvalidSizes to catch this in the debugger.
 ✓ Test testEditingOriginalAddress passed (0.629 seconds)
```

hani-ibrahim:
@bonkey I can't reproduce it :(

bonkey:
For me: `master` / iPhone 7 Plus / 10.2 gives it easily

hani-ibrahim:
still not happening with me 😰 
let's set over it when you return

PR #746, opened 2017-02-17
#727: Fixing collection item size

hani-ibrahim:
fixing #727 

I couldn't get any clue for what could be the problem, but decreasing the item size by 1 point fixes the problem. May be for some reason it needs some margin for approximations however it's not happening on the normal build without tests 😖


hani-ibrahim:
👍

bonkey:
👍

Issue #33, opened 2016-08-03
Add meangingful docs to all public functions


bonkey:
Enabling `missing_docs` in `.swiftlint.yml` would help with this a lot (and it should stay enabled)

bonkey:
Unfinished, yet, pulling back to Ready

PR #748, opened 2017-02-20
#747: Use types named with typealiases instead of pure Strings

bonkey:
#747: Use types named with typealiases instead of pure Strings

### DIFF
--- a/AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
+++ b/AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
@@ -4,45 +4,6 @@
 
 import Foundation
 
-/// Completion closure with `Bool` in `AtlasAPIResult.success` result
-public typealias SuccessCompletion = (AtlasAPIResult<Bool>) -> Void
-
-/// Completion closure with `Customer` in `AtlasAPIResult.success` result
-public typealias CustomerCompletion = (AtlasAPIResult<Customer>) -> Void
-
-/// Completion closure with `Cart` in `AtlasAPIResult.success` result
-public typealias CartCompletion = (AtlasAPIResult<Cart>) -> Void
-
-/// Completion closure with `Checkout` in `AtlasAPIResult.success` result
-public typealias CheckoutCompletion = (AtlasAPIResult<Checkout>) -> Void
-
-/// Completion closure with (`Checkout`, `Cart`) in `AtlasAPIResult.success` result
-public typealias CheckoutCartCompletion = (AtlasAPIResult<(checkout: Checkout, cart: Cart)>) -> Void
-
-/// Completion closure with `Order` in `AtlasAPIResult.success` result
-public typealias OrderCompletion = (AtlasAPIResult<Order>) -> Void
-
-/// Completion closure with `GuestCheckout` in `AtlasAPIResult.success` result
-public typealias GuestCheckoutCompletion = (AtlasAPIResult<GuestCheckout>) -> Void
-
-/// Completion closure with `GuestOrder` in `AtlasAPIResult.success` result
-public typealias GuestOrderCompletion = (AtlasAPIResult<GuestOrder>) -> Void
-
-/// Completion closure with `URL` in `AtlasAPIResult.success` result
-public typealias URLCompletion = (AtlasAPIResult<URL>) -> Void
-
-/// Completion closure with `Article` in `AtlasAPIResult.success` result
-public typealias ArticleCompletion = (AtlasAPIResult<Article>) -> Void
-
-/// Completion closure with `[UserAddress]` in `AtlasAPIResult.success` result
-public typealias AddressesCompletion = (AtlasAPIResult<[UserAddress]>) -> Void
-
-/// Completion closure with `UserAddress` in `AtlasAPIResult.success` result
-public typealias AddressChangeCompletion = (AtlasAPIResult<UserAddress>) -> Void
-
-/// Completion closure with `CheckAddressResponse` in `AtlasAPIResult.success` result
-public typealias AddressCheckCompletion = (AtlasAPIResult<CheckAddressResponse>) -> Void
-
 extension AtlasAPIClient {
 
     public func customer(completion: @escaping CustomerCompletion) {
--- a/AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
+++ b/AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
@@ -74,7 +74,7 @@ extension AtlasAPIClient {
                     return
                 }
 
-                self.createCheckout(fromCardId: cart.id, addresses: addresses) { checkoutResult in
+                self.createCheckout(from: cart.id, addresses: addresses) { checkoutResult in
                     switch checkoutResult {
                     case .failure(let error, _):
                         if self.isCheckoutFailed(error: error) {
@@ -91,20 +91,20 @@ extension AtlasAPIClient {
         }
     }
 
-    public func createCheckout(fromCardId cartId: String, addresses: CheckoutAddresses? = nil, completion: @escaping CheckoutCompletion) {
+    public func createCheckout(from cartId: CartId, addresses: CheckoutAddresses? = nil, completion: @escaping CheckoutCompletion) {
         let parameters = CreateCheckoutRequest(cartId: cartId, addresses: addresses).toJSON()
         let endpoint = CreateCheckoutEndpoint(config: config, parameters: parameters)
         fetch(from: endpoint, completion: completion)
     }
 
-    public func updateCheckout(withId checkoutId: String,
+    public func updateCheckout(with checkoutId: CheckoutId,
                                updateCheckoutRequest: UpdateCheckoutRequest,
                                completion: @escaping CheckoutCompletion) {
         let endpoint = UpdateCheckoutEndpoint(config: config, parameters: updateCheckoutRequest.toJSON(), checkoutId: checkoutId)
         fetch(from: endpoint, completion: completion)
     }
 
-    public func createOrder(fromCheckoutId checkoutId: String, completion: @escaping OrderCompletion) {
+    public func createOrder(from checkoutId: CheckoutId, completion: @escaping OrderCompletion) {
         let parameters = OrderRequest(checkoutId: checkoutId).toJSON()
         let endpoint = CreateOrderEndpoint(config: config, parameters: parameters, checkoutId: checkoutId)
         fetch(from: endpoint, completion: completion)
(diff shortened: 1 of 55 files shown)


bonkey:
👍

AtlasSDK/AtlasSDK/Completions.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  Copyright © 2017 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public typealias AtlasConfigCompletion = (AtlasAPIResult<Config>) -> Void
+
+public typealias AtlasClientCompletion = (AtlasResult<AtlasAPIClient>) -> Void
```
hani-ibrahim: may be documentation for these two?

AtlasSDK/AtlasSDK/Models/Article.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 
 public struct Article {
 
-    public let id: String
+    public let id: SKU
```
hani-ibrahim: `Article.id` -> Config SKU
`Article.unit.it` -> Final SKU

not sure if we should differentiate between them or not 🤔

AtlasSDK/AtlasSDK/Models/GuestPaymentRequest.swift
```diff
@@ -21,12 +21,11 @@ extension GuestPaymentRequest: JSONRepresentable {
         static let metadata = "metadata"
     }
 
-    public func toJSON() -> [String : Any] {
-        var json: [String: Any] = [
-            Keys.method: self.method
+    public func toJSON() -> [String: Any] {
+        return [
+            Keys.method: self.method,
+            Keys.metadata: metadata as Any
```
hani-ibrahim: That's interesting, because we split it over two lines as `metadata` is optional.
what if `metadata` is nil and we are casting it as Any?

hani-ibrahim:
👍

AtlasSDK/AtlasSDK/Models/GuestPaymentRequest.swift
bonkey: Surprisingly, it does work. Checked in a playground before commit.

AtlasSDK/AtlasSDK/Models/Article.swift
bonkey: Yeah, might be better.

However this naming... argh.. how to make it clear for everybody (including internal)

AtlasSDK/AtlasSDK/Completions.swift
bonkey: Sure, it will be. Somebody needs to write it only 👹

hani-ibrahim: yeah I know, it's just because all the file is documented except these two

AtlasSDK/AtlasSDK/Models/Article.swift
bonkey:
```swift
let modelSKU: ModelSKU     = "DU341A00M"
let colorSKU: ColorSKU     = "DU341A00M-102"
let variantSKU: VariantSKU = "DU341A00M-1020375000"
```

hani-ibrahim: uhm, are you sure about the `colorSKU`? may be just `modelSKU` and for the `variantSKU` it could be `fullSKU`, `stockedSKU`, `completeSKU`, `absoluteSKU`

bonkey: But it's not true for `model` because model SKU is **without** color...

I agree, it would be reasonable to name it `basic...` and `complete...`, but it might introduce more confusion when we'll go into color selection, I guess...

bonkey: BTW. we could go further...

```swift
struct ColorSKU {
    fileprivate let internalId: String
}

extension ColorSKU: ExpressibleByStringLiteral {

    init(stringLiteral value: String) {
        self.internalId = value
    }

    init(unicodeScalarLiteral value: String) {
        self.init(stringLiteral: value)
    }

    init(extendedGraphemeClusterLiteral value: String) {
        self.init(stringLiteral: value)
    }

}
```

bonkey: And with that either have 3 types as above (model, color, variant) or enum with 3 cases to return the type of internal data based on some regex pattern :-)

hani-ibrahim: https://media.giphy.com/media/nPZ9zdya9XdKM/giphy.gif

AtlasSDK/AtlasSDK/Completions.swift
bonkey: Actually, I touched this ticket, as it wanted to clarify it before #33...

I'll try hard not to switch to others before #33 is done ;-)

AtlasSDK/AtlasSDK/Models/Article.swift
hani-ibrahim: I don't think we can go with the regex option, as to support the old SKUs.

bonkey: Something in this direction:

```swift
infix operator =~
func =~ (string: String, regex: String) -> Bool {
    return string.range(of: regex, options: .regularExpression) != nil
}

struct SKU {

    enum Kind {
        case model, color, variant, unknown
    }

    fileprivate let internalId: String

    var kind: Kind {
        if internalId ~= "[A-Z0-9]{9}" {
            return .model
        }
        if internalId ~= "[A-Z0-9]{9}-[0-9]{3}" {
            return .color
        }
        if internalId ~= "[A-Z0-9]{9}-[0-9]{10}" {
            return .variant
        }

        return .unknown
    }
}

extension SKU: ExpressibleByStringLiteral {

    init(stringLiteral value: String) {
        self.internalId = value
    }

    init(unicodeScalarLiteral value: String) {
        self.init(stringLiteral: value)
    }

    init(extendedGraphemeClusterLiteral value: String) {
        self.init(stringLiteral: value)
    }

}

extension SKU: CustomStringConvertible {

    var description: String {
        return "\(internalId) (\(kind))"
    }
}

let modelSKU: SKU = "ABC123DEF"
let colorSKU: SKU = "XXYYYXYYX-000"
let variantSKU: SKU = "XXYYYXYYX-0001234567"

dump(modelSKU)
dump(colorSKU)
dump(variantSKU)
```

It's broken, but I must run. if you want – play around with it :-)

hani-ibrahim: but what is the benefit of this? this one is actually the other way around. we allowing the user to enter anything and then the type will be the one that he entered. 

However what we need is the opposite. To say that we need this specific type for this specific method. 

Also I mentioned up, regex is not a good idea as we have old SKUs.

I would prefer simple aliases

AtlasUI/AtlasUI/Utils/AtlasUIClient.swift
```diff
@@ -92,7 +92,7 @@ struct AtlasUIClient {
         }
     }
 
-    static func article(with sku: SKU, completion: @escaping ArticleCompletion) {
+    static func article(with sku: VariantSKU, completion: @escaping ArticleCompletion) {
```
hani-ibrahim: this is `ColorSKU` not `VariantSKU`

AtlasUI/AtlasUI/GetArticleDetailsViewController.swift
```diff
@@ -7,9 +7,9 @@ import AtlasSDK
 
 class GetArticleDetailsViewController: UIViewController {
 
-    let sku: SKU
+    let sku: VariantSKU
 
-    init(sku: SKU) {
+    init(sku: VariantSKU) {
```
hani-ibrahim: this is `ColorSKU` not `VariantSKU`

AtlasUI/AtlasUI/AtlasUI.swift
```diff
@@ -71,13 +71,13 @@ final public class AtlasUI {
     ///   - viewController: The controller in which AtlasUI will be presented over it
     ///   - sku: The SKU for the item that the user want to buy
     /// - Throws: notInitialized is thrown if this method is called before initializing AtlasUI be calling configure(options:completion:)
-    public func presentCheckout(onViewController viewController: UIViewController, forSKU sku: String) throws {
+    public func presentCheckout(onViewController viewController: UIViewController, for sku: String) throws {
```
hani-ibrahim: the SKU is `ColorSKU` not `String`

AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
```diff
@@ -17,10 +17,10 @@ struct GetArticleEndpoint: CatalogEndpoint {
             ])
     }
 
-    let sku: String
+    let sku: VariantSKU
     let fields: [String]?
 
-    init(config: Config, sku: String, fields: [String]? = nil) {
+    init(config: Config, sku: VariantSKU, fields: [String]? = nil) {
```
hani-ibrahim: this is `ColorSKU` not `VariantSKU`

AtlasSDK/AtlasSDK/AtlasAPIClient_public.swift
```diff
@@ -86,7 +86,7 @@ extension AtlasAPIClient {
         fetch(from: endpoint, completion: completion)
     }
 
-    public func article(with sku: SKU, completion: @escaping ArticleCompletion) {
+    public func article(with sku: VariantSKU, completion: @escaping ArticleCompletion) {
```
hani-ibrahim: this is `ColorSKU` not `VariantSKU`

bonkey:
> Apparently we are using ColorSKU for most cases. may be renaming it to BaseSKU ?

Well, I already addressed this "base":

> I agree, it would be reasonable to name it `basic...` and `complete...`, but it might introduce more confusion when we'll go into color selection, I guess...

What I mean by that:

1. Because those SKUs are tricky (as could be seen above, when I already used incorrect ones) it's better to precisely point out what they contain in docs, and then it would be easier to get the context
1. Actually, the whole point of these renames is to make the types clear, not generic to avoid same confusion like with "simple" and "config" naming.

hani-ibrahim:
It's fine for me, but not feeling good with `Color` while we are actually not dealing with colors at all for now. It might introduce confusion while coding at least.
but let's keep it like this

bonkey:
👍

bonkey:
👍

AtlasSDK/AtlasSDK/Types.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  Copyright © 2017 Zalando SE. All rights reserved.
+//
+
+import Foundation
+
+public typealias CartId = String
+public typealias CheckoutId = String
+public typealias CheckoutToken = String
```
hani-ibrahim: What about the login Token?

hani-ibrahim:
👍

bonkey:
👍

bonkey:
👍

PR #755, opened 2017-02-21
How aboout full-fledged SKU types?

bonkey:
PR to #748 to make it more visible and more independent.

And, I'm not sure about it, too.

### DIFF
--- a/AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
+++ b/AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
@@ -17,10 +17,10 @@ struct GetArticleEndpoint: CatalogEndpoint {
             ])
     }
 
-    let sku: ColorSKU
+    let sku: ConfigSKU
     let fields: [String]?
 
-    init(config: Config, sku: ColorSKU, fields: [String]? = nil) {
+    init(config: Config, sku: ConfigSKU, fields: [String]? = nil) {
         self.config = config
         self.sku = sku
         self.fields = fields
--- a/AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
+++ b/AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
@@ -17,10 +17,10 @@ struct GetArticleEndpoint: CatalogEndpoint {
             ])
     }
 
-    let sku: VariantSKU
+    let sku: ColorSKU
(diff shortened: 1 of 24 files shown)


AtlasSDK/AtlasSDKTests/APIGuestOrderTests.swift
```diff
@@ -80,7 +80,7 @@ extension APIGuestOrderTests {
     }
 
     fileprivate var cartRequest: GuestCartRequest {
-        let cartItemRequest = CartItemRequest(sku: "AD541L009-G11", quantity: 1)
+        let cartItemRequest = CartItemRequest(variantSKU: "AD541L009-G11000S000", quantity: 1)!
```
bonkey: @hani-ibrahim This was partially a reason of this PR. I spotted it in the middle of playing with it. Not sure if it's reasonable enough to have it, but maybe it's not that bad, anyway.

bonkey:
👍

AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
```diff
@@ -17,10 +17,10 @@ struct GetArticleEndpoint: CatalogEndpoint {
             ])
     }
 
-    let sku: VariantSKU
+    let sku: ConfigSKU
```
hani-ibrahim: now we have `ConfigSKU` and above it with about 10 lines `Config` 👹

AtlasSDK/AtlasSDK/Models/SKU.swift
```diff
@@ -0,0 +1,104 @@
...
+
+}
+
+extension SKU {
+
+    var isValid: Bool {
+        return Self.pattern.isValid(string: value)
+    }
+
+    init(validate string: String?) throws {
```
hani-ibrahim: probably we don't have any optional SKU anymore?

hani-ibrahim: `validated` or `validate` ? as the output will be a valid SKU or nil

bonkey: So, let's call it `valid`

AtlasSDK/AtlasSDK/Endpoints/GetArticlesEndpoint.swift
bonkey: Don't understand.

Issue #776, opened 2017-03-06
Rename AtlasMockAPI to MockAPI


bonkey:

bonkey:
Sub-task of #85

PR #777, opened 2017-03-06
#776: Rename AtlasMockAPI to MockAPI

bonkey:
It can actually go directly to master, why not. Others probably the same.


bonkey:
👍

bonkey:
👍

hani-ibrahim:
👍

PR #786, opened 2017-03-09
Updated translations

bonkey:
Updated translations


Sources/ZalandoCommerceUI/ZalandoCommerceUI/it.lproj/Localizable.strings
```diff
@@ -110,13 +81,42 @@
 
 /*when happen to have unknown?*/
```
hani-ibrahim: will we keep the comments?

hani-ibrahim:
👍

Sources/ZalandoCommerceUI/ZalandoCommerceUI/it.lproj/Localizable.strings
bonkey: I mostly don't care about them unless they don't contain my CC number ;-)

bonkey:
👍

Issue #775, opened 2017-03-06
Handle images only in media from API


bonkey:
Seems like `media_items` could contain video and 360° images. Show only regular images for now.

PR #792, opened 2017-03-23
#775: handle the new type in media_items 

hani-ibrahim:
fixing #775

### DIFF
--- a/Sources/ZalandoCommerceAPI/ZalandoCommerceAPI/Models/Media.swift
+++ b/Sources/ZalandoCommerceAPI/ZalandoCommerceAPI/Models/Media.swift
@@ -8,6 +8,10 @@ public struct Media {
 
     public let mediaItems: [MediaItem]
 
+    public var firstImage: MediaItem? {
+        return mediaItems.first { $0.itemType == .image }
+    }
+
 }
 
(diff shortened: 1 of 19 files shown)


hani-ibrahim:
👍

Sources/ZalandoCommerceAPI/ZalandoCommerceAPI/Models/Media.swift
```diff
@@ -22,6 +27,15 @@ extension Media {
         public let largeHDURL: URL
     }
 
+    public enum MediaItemType: String {
```
bonkey: If it's inside `Media` it could be just `ItemType`

bonkey:
👍

Sources/ZalandoCommerceAPI/ZalandoCommerceAPI/Models/Media.swift
hani-ibrahim: yes it's inside `Media` but not inside `MediaItem`, so that's why I added explicitly

hani-ibrahim:
👍

bonkey:
👍

